PR #1, opened 2017-09-21
Pull Request to fold in initial fmc_acq430 module support

srobson:


srobson:
Hmm. Not sure how the pulse sim commit ended up here Tom. Too casual a use of the --force argument...

srobson:
Some documentation included and some fixes suggested by Paul

PR #4, opened 2017-12-20
Modifications to Makefiles #3

### LINKED ISSUE #3: Tidy up Makefile structure
Addresses following issues with Makefile/CONFIG file:

- Reference to PandA-rootfs repo (this is done as the make-zpkg script is called from several repos). Problem is that example path given is both obsolete and DLS specific. This file is not managed under version control from any of the calling repos, and the user may choose either not to keep a copy of the rootfs repo accessible, or may keep a stale version of the script. Would it be better to fetch the script from the github when needed with option to check for updates?

- Path to tools. Default path in example script specific to DLS. More generic location would be /opt/Xilinx - symlink to dls_sw? Probably not possible to do in any easy way!

- Licence file location. Again this is specific to DLS and might be better not to expose. Details of the software licencing are specific to the site/tool installation. Would be better for individual users to manage this via bashrc_local scripts

- Default build directory. CONFIG file gives option to modify this, but the location is reset later on back to the default 'build' directory. SlowFPGA build scripts assume knowledge of relative locations of the source and build directories.

glennchid:
* Removed reference to PandA-rootfs repo, and added fetch by wget
* Removed path to licence server
* Fixed to allow build to arbitrary directory
* Closes #3


CONFIG.example
```diff
@@ -35,7 +29,7 @@ SPHINX_BUILD = sphinx-build
 
 # List of default targets build when running make
 #
-MAKE_ZPKG = $(PANDA_ROOTFS)/make-zpkg
+MAKE_ZPKG_URL = https://raw.githubusercontent.com/PandABlocks/PandABlocks-rootfs/master/make-zpkg
```
thomascobb: The original was correct. We should never depend on an internet connection for build, better to specify that the PandABlocks-rootfs repo should be cloned somewhere and use the script from there

Araneidae:
We're not going to merge this as is!  Already discussed with Glenn offline.

PR #5, opened 2018-02-23
Pull request for initial release of ACQ427 support

srobson:
Unsure as to whether you would prefer a pull request to the master branch or the previously massaged for D-TACQ acq430 branch. If this PR is unsuitable please advise and one to the master branch can follow.


common/vhdl/top.vhd
```diff
@@ -88,7 +88,7 @@ port (
     FMC_PRSNT           : in    std_logic;
     FMC_LA_P            : inout std_logic_vector(33 downto 0);
     FMC_LA_N            : inout std_logic_vector(33 downto 0);
-    FMC_CLK0_M2C_P      : in    std_logic;
```
Araneidae: If you're going to change the declaration of a signal, you need to change both parts of it (_N as well as _P)

srobson: We're not using the other half of the diff pair and I wanted to leave it to maintainers to change the others definitions.
As I understand it all CLK_M2C pins are bidirectional in the spec.
Cheers

srobson: Ok. Cracking open the official spec this morning, it seems those lines **are** supposed to be M2C only. We have a looser definition for our extended modules

srobson: The ACQ427 has always been an ELF, yes.

I'll await your feedback then.
Cheers

modules/fmc_acq427/vhdl/fmc_adc_interface.vhd
```diff
@@ -0,0 +1,408 @@
...
+--*************************************************************************************************************************
+
+SEL_CLK_SEL : BUFGMUX
+	port map (O 	=> clk_SEL,
+			I0 	=> clk_PANDA,
+			I1 	=> EXT_CLOCK,
+			S 	=> ADC_CLK_SELECT_REG(0));
+
+--! Sync the ADC_RESET to the selected Clock
+ADC_RESET_RESYNC: process(clk_SEL,ADC_RESET)
```
PaulJG: Clock process only requires the clock in the sensitivity list

modules/fmc_acq427/vhdl/fmc_adc_interface.vhd
```diff
@@ -0,0 +1,408 @@
...
+);
+
+FIFO_DATAIN		<= ADC_DOUT;
+FIFO_DATA_WRITE	<= ADC_FIFOWRITE;
+ADC_DATAOUT		<= ADC_DATAOUT_PCAP;
+
+PCAP_READ_FIFO : process(clk_PANDA)
+begin
+if rising_edge(clk_PANDA) then
+     if ADC_FIFO_COUNT_DATA >= "000001" then
```
PaulJG: Magic number

modules/fmc_acq427/vhdl/fmc_adc_interface.vhd
```diff
@@ -0,0 +1,408 @@
...
+begin
+if Rising_edge(clk_SEL) then
+	CLK_SEL_ADC_RESET	<= ADC_RESET;
+end if;
+end process ADC_RESET_RESYNC;
+
+CLKDIV_COUNTER_RESET <= CLK_SEL_ADC_RESET;
+
+--! This Process Described the Main Divider
+MAINDIV: process (clk_SEL,CLKDIV_COUNTER_RESET,ADCCLK_CLKDIV_DATA,CLKDIV_COUNTER,DIVIDE2)
```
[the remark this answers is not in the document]
Araneidae: Actually, there's a delicate problem here.  I agree that in principle we only need the clock in the sensitivity list ... except, the DIVIDE2 assignment below is *inside* the process.  Whoops.  Take it out of the process and drop the rest of the dependencies and we're good.

modules/fmc_acq427/vhdl/fmc_adc_interface.vhd
```diff
@@ -0,0 +1,408 @@
...
+begin
+DIVIDE2 <=  std_logic_vector(unsigned('0' & ADCCLK_CLKDIV_DATA(15 downto 1)) + 1);
```
Araneidae: Put this outside the process so can drop {{ADCCLK_CLKDIV_DATA}} from the sensitivity list.

Araneidae:
I would like to see the sensitivity lists for purely clocked processes pruned.  However, of course, do be careful that the pruned process really is a pure clocked only process!  Just to be clear, this is what I'm talking about:
1. Pure clocked code:

        process (clk) begin
            if rising_edge(clk) then
               ... clocked code here, nowhere else ...
            end if;
        end process;

2. Clocked code with asynchronous reset (glad not to see any of those)

        process (clk, rst) begin
            if rst = '1' then
                ... asynchronous reset code here
            elsif rising_edge(clk) then
                ... clocked code here, nowhere else ...
            end if;
        end process;

I'm emphasising this point purely because of line 240 of [fmc_adc_interface.vhd](https://github.com/PandABlocks/PandABlocks-FPGA/pull/5/files/f864348855f92abd2b77e64a5bb6f01fd5a851dd#diff-38b7f908001fde2eb7e401ffa57c06fb) which doesn't follow either pattern (code outside the if block).

glennchid:
Verified build of ACQ427 code with Vivado 2015.1.

PR #24, opened 2020-01-30
Make ipmi parsing work for Python3

thomascobb:


Araneidae:
Looks ok.

etc/i2c/ini_file.py
```diff
@@ -34,7 +34,7 @@ def get(self, name, default):
             return default
 
     def __iter__(self):
-        return self.__sections.itervalues()
+        return iter(self.__sections.values())
```
Araneidae: For Python3 only, this would be just
```python
return self.__sections.values()
```

etc/i2c/parse_ipmi.py
```diff
@@ -128,6 +128,7 @@ def next(self, lang_code = None):
             result, self.ix = get_string(self.data, self.ix, lang_code)
             return result
 
+    __next__ = next
```
Araneidae: For Python3 only I guess we'd just rename `next` to `__next__`.

PR #28, opened 2020-03-31
make sure PandABox-fmc_24vio and false path timing constraints for udpontrig are merged

GTHIBAUX:
adding in state on reset of failing APP_NAME = PandABox-fmc_24vio build
+Added false path timing constraints for udpontrig

### DIFF
--- a/modules/fmc_24vio/hdl/max14900_sclk_gen.vhd
+++ b/modules/fmc_24vio/hdl/max14900_sclk_gen.vhd
@@ -68,6 +68,7 @@ begin
             serial_clk <= '0';
             clock_counter <= 0;
             active_o <= '0';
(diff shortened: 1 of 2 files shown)


thomascobb:
Are these the false path constraints that you discussed with @glennchid ?

Araneidae:
I notice there's also a change to a finite state machine in the 24V FMC, not mentioned in the commit message, but the changes look harmless to me.

GTHIBAUX:
Yes the false path constraints was done by @glennchid to make the udpontrig build with vivado 2015.2

The change to a finite state machine in the 24V FMC is to make the build complet. This is safer anyway to have a reset value for FSMs.

PR #29, opened 2020-09-03
Added sign extension to encoders.vhd to be coherent with ssi_sniffer.vhd

valerixb:
Hi
   I have implemented sign extension in the encoders.vhd to see negative numbers when we have encoders with less than 32 bits; this makes it coherent with the sign extension code in ssi_sniffer.vhd

### DIFF
--- a/targets/PandABox/hdl/encoders.vhd
+++ b/targets/PandABox/hdl/encoders.vhd
@@ -235,7 +235,9 @@ begin
         lp_test: for i in 31 downto 0 loop
            -- Discard bits not being used and MSB and LSB and append zeros on to top bits
            if (i > 31 - bits_not_used - unsigned(MSB_DISCARD_i) - unsigned(LSB_DISCARD_i)) then
-               posn_o(i) <= '0';
+               --posn_o(i) <= '0';
+               -- sign extension
+               posn_o(i) <= posn(31 - to_integer(bits_not_used - unsigned(MSB_DISCARD_i)));
            -- Add the LSB_DISCARD on to posn index count and start there
            else
                posn_o(i) <= posn(i + to_integer(unsigned(LSB_DISCARD_i)));
--- a/targets/PandABox/hdl/encoders.vhd
+++ b/targets/PandABox/hdl/encoders.vhd
@@ -233,7 +233,7 @@ begin
         -- BITS not begin used
         bits_not_used <= 31 - (unsigned(INENC_BITS_i(4 downto 0))-1);
         lp_test: for i in 31 downto 0 loop
-           -- Discard bits not being used and MSB and LSB and append zeros on to top bits
+           -- Discard bits not being used and MSB and LSB and extend the sign
            if (i > 31 - bits_not_used - unsigned(MSB_DISCARD_i) - unsigned(LSB_DISCARD_i)) then
                --posn_o(i) <= '0';
                -- sign extension
@@ -546,5 +546,4 @@ clkin_filt : entity work.delay_filter port map (
     filt_o  => CLK_IN
 );
 end rtl;
-
 


PaulJG:
If you are sign extending then this would be better
posn_o(i) <= posn(31);
instead of this
posn_o(i) <= posn(31 - to_integer(bits_not_used - unsigned(MSB_DISCARD_i)));

PaulJG:
If you are sign extending then this would be better
posn_o(i) <= posn(31);
instead of this
posn_o(i) <= posn(31 - to_integer(bits_not_used - unsigned(MSB_DISCARD_i)));

valerixb:
I am no expert, so maybe I am getting this wrong, but I was thinking at one exmple: suppose we have a 16 bit multiturn encoder but we are interested only in the single turn data, which are encoded in the 8 least significant bits: in this case, we expect to get a number between -128 and +127. Now suppose that we start from 0 and we move in the negative direction just over 180 degrees, so the 16-bit value is -129; this value, sign extended to 32 bit by ssi_sniffer.vhd still reads -129 and in binary it's: 11111111_11111111_11111111_01111111.
If encoders.vhd sign-extends with 
posn_o(i) <= posn(31); 
we still have -129, which is out of range, as we wanted something between -128 and +127. If the sign extension is done with 
posn_o(i) <= posn(31 - to_integer(bits_not_used - unsigned(MSB_DISCARD_i)));
then the leading 0 is extended, we get 00000000_00000000_00000000_01111111 , which is +127, which is what we expect from the wrapping.
Do you agree or is there something that I am missing?
Thank you! Best regards- Valerio Bassetti

glennchid:
Hi Valerio, 

Do you mean
`posn_o(i) <= posn(31 - to_integer(bits_not_used + unsigned(MSB_DISCARD_i)));`
rather than 
`posn_o(i) <= posn(31 - to_integer(bits_not_used - unsigned(MSB_DISCARD_i)));`?

Would it not be easier though to simply slice the vector and resize it in signed representation, which would remove the need for the loop? For example, something like:

```
ps_select : process(clk_i)  
    variable intBITS, intMSB_DIS, intLSB_DIS : integer;  
    variable posn_tmp : signed(31 downto 0);  
  begin
    if rising_edge(clk_i) then
      intBITS     := to_integer(unsigned(INENC_BITS_i));
      intMSB_DIS  := to_integer(unsigned(MSB_DISCARD_i));
      intLSB_DIS  := to_integer(unsigned(LSB_DISCARD_i));
      posn_tmp    := resize(signed(posn((intBITS - intMSB_DIS - 1) downto intLSB_DIS)), 32);
      posn_o      <= std_logic_vector(posn_tmp);
  end if; 
end process;
```

Would this work?

valerixb:
Hi Christian
Yes, thank you: I forgot to flip the sign when I added the parenthesis.
As far as the alternative coding with the signed resize is concerned, I am no real expert in VHDL, so I really don't have a strong opinion on that, other than agreeing that your solution looks cleaner and clearer to me; please go ahead with the one you think is better. Actually, if the second way is chosen, maybe it's worth modifying also the sign extension that is performed at lower level by the biss/ssi sniffers and masters, to have a common solution? I am thinking of biss_sniffer.vhd line 310, biss_master.vhd line 306, ssi_sniffer.vhd line 177 and ssi_master.vhd line 141. Sorry, but I am new to github and I don't know how to put hyperlinks in the text.
Thank you! Best regards
     Valerio

glennchid:
Hi Valerio, ah, unfortunately the vector slicing approach does not work in synthesis - I had overlooked the fact that the slice indices are not constants! Pity, as it was more compact and easier to understand - but I can see now why the loop had to be used.

@tomtrafford or @PaulJG - please could you merge this if you are happy to, and take into account the sign flip noted above?

Best wishes, Glenn.

targets/PandABox/hdl/encoders.vhd
```diff
@@ -235,7 +235,9 @@ begin
         lp_test: for i in 31 downto 0 loop
            -- Discard bits not being used and MSB and LSB and append zeros on to top bits
            if (i > 31 - bits_not_used - unsigned(MSB_DISCARD_i) - unsigned(LSB_DISCARD_i)) then
-               posn_o(i) <= '0';
+               --posn_o(i) <= '0';
+               -- sign extension
+               posn_o(i) <= posn(31 - to_integer(bits_not_used - unsigned(MSB_DISCARD_i)));
```
tomtrafford, line 240:
```suggestion
               posn_o(i) <= posn(31 - to_integer(bits_not_used + unsigned(MSB_DISCARD_i)));
```

tomtrafford (changes requested):
Hi Valerio, if you're happy with this change that Glenn had suggested I will merge this.

Thanks, Tom

targets/PandABox/hdl/encoders.vhd
```diff
@@ -235,7 +235,9 @@ begin
         lp_test: for i in 31 downto 0 loop
            -- Discard bits not being used and MSB and LSB and append zeros on to top bits
```
valerixb, line 236: I think we should also change the comment to:
-- Discard bits not being used and MSB and LSB and extend the sign

targets/PandABox/hdl/encoders.vhd
valerixb: OK: agree; thanks!

targets/PandABox/hdl/encoders.vhd
tomtrafford:
```suggestion
           -- Discard bits not being used and MSB and LSB and extend the sign
```

tomtrafford (changes requested):

PaulJG:
Hi All,

I did the loop code and I used a loop so as synth wouldn’t complain about the signal or variable not being a fixed value.
I think that all the code needs is a comment saying why it was done that way even giving any example, because as soon as I read the email you sent I realised why you did that.

PR #30, opened 2020-09-03
added gray encoding options to INENC and OUTENC

valerixb:
Hi
    in our facility (MaxIV Laboratory, Lund) we have gray-encoded absolute encoders so I added a parameter to INENC and OUTENC to choose between straight binary and gray encoding.

### DIFF
--- a/common/hdl/defines/support.vhd
+++ b/common/hdl/defines/support.vhd
@@ -4,6 +4,9 @@ use ieee.numeric_std.all;
 
 package support is
 
+constant c_BINARY_ENCODING  : std_logic_vector(0 downto 0) := "0";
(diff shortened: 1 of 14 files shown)


GTHIBAUX:
We have tested with the following absolut encoders with succes:
kubler, bissc 12 bit ST + 12 MT > 24 bits binary
Heidenhain , ROC413 // SSI 13 bits Gray

As expected relative coders are not impacted by INENC.ENCODING OUTENC.ENCODING
Maybe we could add in ENCODING descrition that this param only apply to relative encoders set in PROTOCOL?

thomascobb:
Closed in favour of a new PR that Valerio will open

PR #31, opened 2020-09-10
warn instead of error if FMC zpkg installed but no FMC detected

blondejamtart:


etc/check_ipmi
```diff
@@ -27,6 +27,13 @@ if eeprom_key == 'ignore':
 try:
     allow_16bit = eeprom_key == '16-bit'
     image = eeprom.read_eeprom(allow_16bit)
+except OSError as e:
+    if e.errno == 6: # No FMC detected
```
Araneidae, line 31: `errno == 6` means "no FMC detected"?  Really?!

I suppose I'd be slightly happier with the test `e.errno == errno.ENXIO`.  At least the manpage errno(3) tells us that this means:

> No such device or address (POSIX.1-2001).

which I guess is a pretty fair diagnosis.

Araneidae (changes requested):
Ok, change the 6 to `errno.ENXIO` and add a comment about "no such device" and I guess this is fine.

I do have one question though, and I'll flag @thomascobb for this: is it fundamentally sound to simply load the FPGA image when the FMC is missing?  I accept that it is probably harmless, unless the "no such device" is simply that the EEPROM is broken or otherwise not responding, but who is going to see the error message and take any action.

If the FPGA image expects an FMC and the FMC is not present ... what's the point of loading the image?  If it's for some special test case then we should have an explicit override mechanism instead; I don't see how this improves the experience of users.

So.  Why are we doing this?

thomascobb:
> I do have one question though, and I'll flag @thomascobb for this: is it fundamentally sound to simply load the FPGA image when the FMC is missing? I accept that it is probably harmless, unless the "no such device" is simply that the EEPROM is broken or otherwise not responding, but who is going to see the error message and take any action.

Good point. This was raised by Soleil as they left a loopback firmware in, then removed the FMC card, then the TCP server wouldn't start with a cryptic error. I guess we have two options to improve user experience:

1) error, but with a more useful error message telling the user that PandA won't boot until the FMC is present
2) warn, then rely on the user to look at the FMC_PRSNT field on the block

The first is a clear flag something is broken, and is safe against a broken EEPROM, but you can't see this from EPICS/TANGO.

The second means the error is reported in the control system only, and might be missed, but the system will boot allowing you to poke it more easily.

I think you've persauded me that the first is actually better.

@blondejamtart please could you change it to `sys.exit(1)`

etc/check_ipmi
```diff
@@ -27,6 +28,13 @@ if eeprom_key == 'ignore':
 try:
     allow_16bit = eeprom_key == '16-bit'
     image = eeprom.read_eeprom(allow_16bit)
+except OSError as e:
+    if e.errno == errno.ENXIO:  # No such device or address (i.e. no FMC EEPROM detected)
+        print('FMC was expected but none detected')
+        sys.exit(1)
+    else:
+        print('Unable to read FMC EEPROM:', e)
```
Araneidae, line 36: Oh, that's irritating.  Just noticed that lines 36-37 repeat lines 38-39.  I really can't think of a decent workaround, so meh.

Araneidae (approved):

PR #32, opened 2020-09-14
Fix SlowFPGA compilation issue to take account the BUILD_DIR defined …

shu-soleil:
One FPGA app compilation was failed due to a different BUILD_DIR defined to the default one (top/build) in CONFIG file. 

The vhdl working paths are currently fixed to “top/xxx” in “PandA-FPGA/targets/PandABox/SlowFPGA/Makefile” file @line 52.

The workaround solution is to use default BUILD_DIR. Otherwise “slow_top.files.jinjia2” might need to take account the BUILD_DIR defined in CONFIG file and “PandABox-FPGA/targets/PandABox/SlowFPGA/Makefile” not to impose “top/” to "$(LIST_FILE): $(AUTOGEN)/slow_top.files".

### DIFF
--- a/common/python/generate_app.py
+++ b/common/python/generate_app.py
@@ -201,7 +201,7 @@ class AppGenerator(object):
             "descriptions.jinja2", context, config_dir, "description")
         self.expand_template(
             "sim_server.jinja2", context, self.app_build_dir, "sim_server")
-        context = jinja_context(app=self.app_name)
+        context = jinja_context(app_dir=self.app_build_dir)
         self.expand_template(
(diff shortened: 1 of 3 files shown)


glennchid:
Hi Shu, could you compare these changes to the changes that I was proposing to the slow FPGA build in the 'target' branch? Trying to build to an arbitrary BUILD_DIR location has never worked due to BUILD_DIR being hard-coded in other scripts. I had fixed this in the 'target' branch, but other issues meant that the branch was not ready to pull into master.

PR #33, opened 2020-10-12
numerous changes and restructuring for target independent build

glennchid:
This branch contains many changes to the build system for target independence including generation of target specific FSBL and devicetree from within the FPGA repo, using 'make boot'. It also allows build to an arbitrary build location (this did not work previously).

There are several items to note/ things still to do:

- Currently fails 'python_tests' and 'hdl_test'

- Hardware blocks, previously under 'targets/<target>/' now moved to 'modules' directory. I made some hacky changes to the paths in several python files to make this work (and left the previous lines commented out). There may be a neater way to make these changes; also we now have directories for inenc, outenc (needed by the python code), and 'encoders' (needed for the hdl) - this need to be tidied up.

- Encoder and TTL blocks now have dependencies on the slow FPGA (which is target specific). 

- Several area could do with more jinja templating: for example 'load-panda_firmware' script, which also loads the slow FPGA firmware.

- IP build: we currently build all IP for a given target regardless of the app being built. Alternatively, we could the build the IP as and when required but this would require make to have knowledge of the target/app. At the moment there is a clear separation between the 'make' system and the autogenerated parts.


glennchid:
Issue with encoder and ttlin block depending on slow_fpga has been addressed by passing the relevant parameters (ENC_MODE, and TTL_TERM) and their respective WSTBs from the encoder block to the system block, where the tlp packet for the Spartan is generated. This should be checked with the hardware to verify it still functions correctly.

glennchid:
Now passing python tests and hdl_timing build

glennchid:
Now passing all tests: python_tests, python_timing & hdl_tests

glennchid:
Moved slow_load across from server repo.

Also, fixed issue with zynq and slow FPGA builds always being treated as phony targets (due to having phony dependency - 'VERSION' containing the git hash). Now only update FPGA targets if not run previously, git hash differs or contains the 'dirty' string. I am happy with how it works, but the tests/comparisons could probably be done in a neater way in make, rather than relying on the shell for this. @Araneidae and @thomascobb - can you please review (PREV_VERSION target in top-level Makefile)?

thomascobb (commented):
I think the general concept is reasonable, I don't know of any tricks that would accomplish this any other way. Maybe some naming changes might make it more readable though?

- PREV_VER -> VERSION_FILE
- PREV_VERSION -> update_version_file

Araneidae:
I have converted this to draft, as it will need to be merged at the same time as [PandABlocks-rootfs#18](https://github.com/PandABlocks/PandABlocks-rootfs/pull/18).

Araneidae:
Looks like master needs to be merged back into this branch first!

Araneidae:
Having looked at [PandABlocks-server PR#10](https://github.com/PandABlocks/PandABlocks-server/pull/10) and after our discussions about the build difficulties I am persuaded that we should drop `slow_load` from this merge and leave it in the server.  This would mean that PandABlocks-FPGA is not building any code that runs on the target rootfs, which solves some thorny problems.

glennchid:
> Looks like master needs to be merged back into this branch first!

The merge with 'master' has been done on the 'target2020' branch, where I am working on the update for Vivado 2020. This is just waiting on the installation of Vivado 2020.2 at DLS, then I will merge 'target2020' into this PR.

glennchid:
Upgrade to Vivado 2020.2. 

All tests pass and zpackages build okay, minimally tested on PandABox and ZedBoard targets.

Issue discovered with PCAP not re-arming correctly, however this problem is also present in the destination branch (master @ v2.2a1).

PR #35, opened 2020-10-14
WIP: Feature/write eeprom

blondejamtart:
Write 24v FMC EEPROM
- fixes for more robust detection of 16-bit EEPROMs


thomascobb:
Please can you extract all the write functions and put them in a separate standalone "write_24vio_eeprom.py" script. We shouldn't install it, it can go in this repo and we'll scp it across when we need it. This means we don't need to detect 16-bit mode, it should be a parameter to the script.

I think the "reading the header" detection is probably ok, it's certainly the simplest one to understand.

etc/i2c/eeprom.py
```diff
@@ -64,8 +71,11 @@ def read_8bit_address(device = 0x50, length = 256):
 
 
 def read_eeprom(allow_16bit):
-    if allow_16bit and detect_16bit():
-        eeprom = read_16bit_address()
+    if allow_16bit:
+        if detect_16bit():
+            eeprom = read_16bit_address()
+        else:
+            eeprom = read_16bit_address()
```
thomascobb, line 78: This line is the same as 2 above...

etc/i2c/eeprom.py
```diff
@@ -18,18 +19,24 @@ def read_bytes(bus, device, length = 32):
...
+# NOTE: this fails for an 8-bit device which doesn't contain valid data
+
 def detect_16bit(device = 0x50):
-    # Read two half pages.  If they are the same then we can assume that we were
-    # able to reset
-    first_read  = read_8bit_address(device, length = 15)
-    second_read = read_8bit_address(device, length = 15)
-    return first_read != second_read
+    first_read  = read_8bit_address(device, length = 8)
+    second_read = read_8bit_address(device, length = 8)
```
thomascobb, line 29: I might be missing something here, but doesn't this require the address counter to be at the start like the old code did?

etc/write_eeprom.py
```diff
@@ -0,0 +1,55 @@
...
+            except OSError as e:
+                if not e.errno == errno.ENXIO:  # expected error is no ACK (device not ready for next write)
+                    raise e
+        assert (len(readback) == 1 and b == readback[0]), "readback data does not match"
+        readback = None
+        offset += 1
+
+
+if __name__ == "__main__":
+    byte_data = stdin.buffer.read()
```
thomascobb, line 48: Can you put the EEPROM data in this file? Did you generate it from a library?

etc/write_eeprom.py
```diff
@@ -0,0 +1,55 @@
...
+
+    bus = smbus2.SMBus(0)
+    if eeprom.detect_16bit():
```
thomascobb, line 51: Can we add an argument to write as 8-bit rather than detect please?

thomascobb (changes requested):

etc/i2c/eeprom.py
blondejamtart: read_8bit_address attempts to read data starting from zero (sends message to set address to zero). We attempt to read sending an 8bit address twice (if the counter was at zero, the first read would return the header but the second wouldn't), and if we don't get the header data twice we assume it didn't accept the address send and hence is 16bit. We then move on to check with a 16-bit read, and error if we still don't get a valid header.

etc/write_eeprom.py
blondejamtart: generated with the dtacq tool (https://github.com/petermilne/fmctools/blob/master/tools/fru-generator).
Will put it in once I've generated a proper one with correct part number etc. I guess I should remove the option to pipe in via stdin?

etc/write_eeprom.py
blondejamtart: will do. do we have argparse on the PandA or should I just use sys.argv?

etc/i2c/eeprom.py
thomascobb: I see, that makes sense

etc/write_eeprom.py
thomascobb: Yes, it's part of the standard library

etc/i2c/eeprom.py
```diff
@@ -18,18 +19,24 @@ def read_bytes(bus, device, length = 32):
...
+    try:
+        parse_ipmi.parse_header(first_read)
+        parse_ipmi.parse_header(second_read)
+    except AssertionError:
```
Araneidae, line 33: Not happy about specifically catching `AssertionError`.  I'd suggest either catching all exceptions (if that's what you want), or hunting down the offending `assert` statements and replacing them with raises of an appropriate new `Exception` subclass.

etc/i2c/eeprom.py
```diff
@@ -18,18 +19,24 @@ def read_bytes(bus, device, length = 32):
...
+# If not we assume the device is 16-bit addressed 
+# NOTE: this fails for an 8-bit device which doesn't contain valid data
+
 def detect_16bit(device = 0x50):
-    # Read two half pages.  If they are the same then we can assume that we were
-    # able to reset
-    first_read  = read_8bit_address(device, length = 15)
-    second_read = read_8bit_address(device, length = 15)
-    return first_read != second_read
+    first_read  = read_8bit_address(device, length = 8)
```
Araneidae, line 28: Why the change in length from 15 to 8?  Otherwise there is no diff here

etc/i2c/eeprom.py
```diff
@@ -64,8 +71,13 @@ def read_8bit_address(device = 0x50, length = 256):
...
 def read_eeprom(allow_16bit):
-    if allow_16bit and detect_16bit():
-        eeprom = read_16bit_address()
+    if allow_16bit:
+        if detect_16bit():
+            eeprom = read_16bit_address()
+        else:
+            print("Warning: 16 bit set but not detected")
+            if allow_16bit > 1:
+                eeprom = read_16bit_address()
```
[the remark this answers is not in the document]
blondejamtart: I think github has managed to show you an outdated version of the files here....this changes has been reverted now

etc/panda-fpga.list
```diff
@@ -20,6 +20,7 @@ t share/panda-fpga etc/load-panda-firmware
 T lib/python/site-packages/i2c etc/i2c/*.py
 t bin etc/check_ipmi
 t bin etc/show_ipmi
+t bin etc/write_ipmi
```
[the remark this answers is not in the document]
blondejamtart: outdated file again (this has gone in more recent commits)

etc/i2c/eeprom.py
blondejamtart: the lines lower down in the function have changed...the parse header function expects length 8

etc/i2c/eeprom.py
blondejamtart: the asserts are inside parse_header....maybe a ParsingException?

etc/panda-fpga.list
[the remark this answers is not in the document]
Araneidae: Sorry about the outdated comments: think I started a review yesterday and you made changes before I'd finished!

Araneidae (approved):

blondejamtart:
Maybe I misinterpreted @thomascobb's request to "put the EEPROM data in this file". 
I generated this with the dtacq tool (https://github.com/petermilne/fmctools/blob/master/tools/fru-generator) since it was mentioned in the JIRA ticket for this, but the create_ipmi script is probably fine and makes more sense to use.
Maybe we should get write_eeprom to use create_ipmi to create the relevant image from an input ini and then write it.
Do we think this script should go into another repo? If so which one?

thomascobb:
> Maybe I misinterpreted @thomascobb's request to "put the EEPROM data in this file".

Nope, you didn't misinterpret it, I suggested it go in this repo. Not so sure now...

> I generated this with the dtacq tool (https://github.com/petermilne/fmctools/blob/master/tools/fru-generator) since it was mentioned in the JIRA ticket for this, but the create_ipmi script is probably fine and makes more sense to use.

I didn't know create_ipmi existed, that's much better

> Maybe we should get write_eeprom to use create_ipmi to create the relevant image from an input ini and then write it.

Sounds like a good idea.

> Do we think this script should go into another repo? If so which one?

I'm not sure. We could either keep all this stuff together in the FPGA repo, the ini file and the write_eeprom and create_ipmi, or move them all to the PandABox-tst repo.

@Araneidae what do you reckon? Is create_ipmi used in the FPGA repo at all?

Araneidae:
Maybe place this stuff in a separate `tools` directory?  The `etc` directory is meant to contain stuff that's part of the installation.

thomascobb:
sounds good

Araneidae:
Ideally the actually IPMI image definition (the .ini file) would live in the 24V FMC repo, but I guess that's not particularly practical right now, so it can live in a special place here.

modules/fmc_24vio/ipmi_definition.ini
```diff
@@ -0,0 +1,73 @@
+[Board]
```
blondejamtart, line 1: @thomascobb  where would I find the information needed to populate this properly?

etc/i2c/eeprom.py
thomascobb: Could you put this text in as a comment please?

modules/fmc_24vio/ipmi_definition.ini
thomascobb: @amcdls  should have the info...

thomascobb:
@blondejamtart what is the status of this?

tomtrafford:
Can we run the 2.X zpkg when the Eeprom is already written?

modules/fmc_24vio/ipmi_definition.ini
```diff
@@ -0,0 +1,73 @@
...
+manufacture date: 2011-12-08 00:00
+manufacturer: DLS Ltd
+product name: 24V GPIO
+serial number: 36622945
```
tomtrafford, line 5: Should be a random number - not hard coded

EmilioPeJu:
Just let you know that I created PR https://github.com/PandABlocks/PandABlocks-FPGA/pull/153 which is an improvement of this branch

Issue #34, opened 2020-10-13
position_sign_extension is not optionnal


GTHIBAUX:
with Src version: PandA SW: 2.0.1 FPGA: 2.0.3C22 04bb2b28 04bb2b28 rootfs: PandA 2.0.1

We have tested with these relative encoders:
kubler,  bissc 12 bit ST + 12 MT > 24 bits (BINARY)
baumer,  BMMH 42S1G05C12/13B25 (seems to be BINARY with 1 sign bit)

The value returned is always signed and this could induce differences in value from the motor controler.
And majority of our relative encoders are BINARY (not signed).

Maybe we should add SIGNED enum in IN/OUTENC. ENCODING as done with encoder_gray branch?

thomascobb:
That sounds like a good idea, only sign extend if SIGNED=Yes

valerixb:
Hi
    maybe we should add another property like, for example, IN/OUTENC.SIGN that can be SIGNED or UNSIGNED? This allows to use different encodings (binary, gray) along with different sign choices (signed, unsigned) and all 4 the possibilities can be selected. Or maybe directly put 4 possibilities in the IN/OUTENC.ENCODING property.

thomascobb:
This raises an interesting question, how do we handle signed vs unsigned values at the moment? Most of the time they are passed straight through, but the pcap code treats them as signed:

https://github.com/PandABlocks/PandABlocks-FPGA/blob/4bb2b283ecf7b61503a3126a49714d5240d0d47c/targets/PandABox/hdl/pcap_frame_mode.vhd#L128

So this means that all positions are treated as 32-bit signed to PCAP. This means that 32-bit unsigned encoders won't be represented correctly in PCAP. We can't really do much about this as the averaging code is already pretty expensive.

I think I slightly prefer putting all the options into ENCODING, but if it's easier to have 2 fields then that is fine too.

thomascobb:
Closed via #39

PR #39, opened 2021-02-10
Encoder gray + signed/unsigned

valerixb:
Possibility to choose between gray and binary encoding and signed/unsigned value for absolute encoders; default is unsigned binary. Description has been updated as per GTHIBAUX comment on PR #30, to show that this affects absolute and not incremental encoders. This pull request also solves issue #34 (position_sign_extension is not optional).

### DIFF
--- a/common/hdl/defines/support.vhd
+++ b/common/hdl/defines/support.vhd
@@ -4,6 +4,9 @@ use ieee.numeric_std.all;
 
 package support is
 
+constant c_BINARY_ENCODING  : std_logic_vector(0 downto 0) := "0";
+constant c_GRAY_ENCODING    : std_logic_vector(0 downto 0) := "1";
+
 --
 -- Functions
 --
--- a/common/hdl/defines/support.vhd
+++ b/common/hdl/defines/support.vhd
@@ -4,8 +4,10 @@ use ieee.numeric_std.all;
 
 package support is
 
-constant c_BINARY_ENCODING  : std_logic_vector(0 downto 0) := "0";
-constant c_GRAY_ENCODING    : std_logic_vector(0 downto 0) := "1";
+constant c_UNSIGNED_BINARY_ENCODING  : std_logic_vector(1 downto 0) := "00";
+constant c_UNSIGNED_GRAY_ENCODING    : std_logic_vector(1 downto 0) := "01";
+constant c_SIGNED_BINARY_ENCODING    : std_logic_vector(1 downto 0) := "10";
+constant c_SIGNED_GRAY_ENCODING      : std_logic_vector(1 downto 0) := "11";
 
 --
 -- Functions
--- a/common/hdl/shifter_in.vhd
+++ b/common/hdl/shifter_in.vhd
@@ -66,12 +66,12 @@ begin
                 smpl_hold <= (others => '0');
             -- Shift data when enabled.
             elsif (enable_i = '1') then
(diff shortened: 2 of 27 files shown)


common/hdl/shifter_in.vhd
```diff
@@ -65,8 +66,12 @@ begin
                 smpl_hold <= (others => '0');
             -- Shift data when enabled.
             elsif (enable_i = '1') then
-                if (clock_i = '1') then
-                    smpl_hold <= smpl_hold(DW-2 downto 0) & data_i;
+				if (clock_i = '1') then
```
Araneidae, line 69: I notice that this commit introduces tabs that were not previously present in the source.

I suspect this may be a widespread issue.  @thomascobb , it might be worth have a look at the code base to flush out any creeping tab characters.

Contributors: please configure your editor appropriately!

valerixb: Sorry: rookie mistake. I fixed all the extra TABs, leaving only those in makefiles, device tree files (dts) and .prn files; I committed the changes to https://github.com/valerixb/PandABlocks-FPGA/tree/encoder_gray. 
I am no real expert of git: what do you think is the best way to bring those changes here?
Thank you; best regards

common/hdl/defines/support.vhd
```diff
@@ -4,6 +4,11 @@ use ieee.numeric_std.all;
 
 package support is
 
+constant c_UNSIGNED_BINARY_ENCODING  : std_logic_vector(1 downto 0) := "00";
```
thomascobb, line 7: I think that the existing behaviour is `SIGNED_BINARY_ENCODING`. If this is correct, please can we assign `00` to that, so that an upgrade will be backwards compatible? If we merge as is, then we will break anyone with a 32-bit signed encoder...

targets/PandABox/blocks/inenc/inenc.block.ini
```diff
@@ -14,6 +14,14 @@ description: Type of absolute/incremental protocol
 2: BISS
 3: enDat
 
+[ENCODING]
+type: param enum
+description: Position encoding (for absolute encoders)
+0: Unsigned Binary
```
thomascobb, line 20: Change would also need to be made here

targets/PandABox/blocks/outenc/outenc.block.ini
```diff
@@ -38,6 +38,14 @@ description: Type of absolute/incremental protocol
 4: ABZ Passthrough
 5: DATA Passthrough
 
+[ENCODING]
+type: param enum
+description: Position encoding (for absolute encoders)
+0: Unsigned Binary
```
thomascobb, line 44: And here

thomascobb (changes requested):

common/hdl/defines/support.vhd
valerixb: Tom, it was an explicit request from Soleil: as in the original panda code there was no sign extension (I think it was a bug), using "unsigned binary" as default (00) makes the change backwards compatible. If we put "signed binary" we brake compatibility with all FW before #29 (september 2020), which I think is the bulk of the Pandas. What do you think?

thomascobb: ok, that makes sense, I will merge this as is

thomascobb (approved):

PR #42, opened 2021-03-10
Support doing nothing on blank table

### LINKED ISSUE #38: SEQ block: empty table but trig
**the sequencer generates a sequence while the table is empty.** 

### senario:
- load table sequence
- activate the sequencer
- Deactivate the sequencer
- put an empty table.
- activate the sequencer
-> a part of the sequence is generated.

thomascobb:
Fixes #38

### DIFF
--- a/modules/seq/hdl/seq.vhd
+++ b/modules/seq/hdl/seq.vhd
@@ -89,8 +89,6 @@ constant c_state_wait_trigger   : std_logic_vector(2 downto 0) := "010";
 constant c_state_phase1         : std_logic_vector(2 downto 0) := "011";
 constant c_state_phase2         : std_logic_vector(2 downto 0) := "100";
 
-signal reset_i              : std_logic := '0';
-
 signal TABLE_FRAMES         : std_logic_vector(15 downto 0);
 
 signal current_frame        : seq_t;
@@ -276,7 +274,7 @@ enable_mem_reset <= '1' when (current_frame.time1 = x"00000000") else '0';
 
 
 load_next <= '1' when (enable_rise = '1' and  seq_sm = WAIT_ENABLE) or ( seq_sm = PHASE_2  and presc_ce = '1' and
-                      tframe_counter = next_ts -1 and last_table_repeat = '0' and last_line_repeat = '1') or 
+                      tframe_counter = next_ts -1 and last_table_repeat = '0' and last_line_repeat = '1') or
                       (current_frame = next_frame and seq_sm = PHASE_1 and unsigned(PRESCALE) < to_unsigned(2,32)
                       and (current_frame.time2 = to_unsigned(0,32) or current_frame.time2 = to_unsigned(1,32)))
                       else '0';
@@ -289,15 +287,7 @@ if rising_edge(clk_i) then
     --
 --    load_next <= '0';
 
-    -- Reset all registers and state machine.
-    if (reset_i = '1') then
-        seq_sm <= WAIT_ENABLE;
-        out_val <= (others => '0');
-        active <= '0';
-        LINE_REPEAT_o <= (others => '0');
-        TABLE_LINE_o <= (others => '0');
-        TABLE_REPEAT_o <= (others => '0');
-    elsif (TABLE_START_WSTB = '1') then
+    if (TABLE_START_WSTB = '1') then
         out_val <= (others => '0');
         seq_sm <= LOAD_TABLE;
     elsif (enable_fall = '1' and enable_i = '0') then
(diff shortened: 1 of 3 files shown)


glennchid (approved):
This all looks okay to me.

modules/seq/hdl/seq.vhd
```diff
@@ -315,7 +305,11 @@ if rising_edge(clk_i) then
             when WAIT_ENABLE =>
                 -- TABLE load_started
                 reset_table <= '0';
-                if enable_rise = '1' then
+                if unsigned(TABLE_LENGTH) = to_unsigned(0,32) then
```
Araneidae, line 308: You can compare an `unsigned` with 0.  Is there a good reason why `TABLE_LENGTH` isn't already an `unsigned` at this point?

modules/seq/hdl/seq.vhd
```diff
@@ -315,7 +305,11 @@ if rising_edge(clk_i) then
...
+                    TABLE_REPEAT_o <= to_unsigned(0,32);
+                    TABLE_LINE_o <= to_unsigned(0,16);
+                    LINE_REPEAT_o <= to_unsigned(0,32);
```
Araneidae, line 311: It's more idiomatic to write `(others => '0')` rather than `to_unsigned(0,n)`.

modules/seq/hdl/seq.vhd
```diff
@@ -289,15 +287,7 @@ if rising_edge(clk_i) then
...
 --    load_next <= '0';
 
-    -- Reset all registers and state machine.
-    if (reset_i = '1') then
-        seq_sm <= WAIT_ENABLE;
-        out_val <= (others => '0');
-        active <= '0';
-        LINE_REPEAT_o <= (others => '0');
-        TABLE_LINE_o <= (others => '0');
-        TABLE_REPEAT_o <= (others => '0');
```
Araneidae, line 299: Are these variables now statically initialised elsewhere (or do we know their initial values don't matter)?  I welcome losing the `reset_i`.  If `reset_i` is known to be _always_ set to `'0'` then this comment can be ignored.

modules/seq/hdl/seq.vhd
```diff
@@ -490,17 +484,13 @@ port map (
 process(clk_i)
 begin
     if rising_edge(clk_i) then
-        if (reset_i = '1') then
+        if (presc_reset = '1') then
```
Araneidae, line 487: Any chance we can consistently lose the extra brackets on the `if` statements?  They're not part of VHDL syntax, and the coding style in this file is inconsistent.

Araneidae (commented):
Some style comments on the vhdl.

modules/seq/hdl/seq.vhd
thomascobb: No good reason, I don't know enough VHDL to know what the alternative could be, can you make a suggestion?

modules/seq/hdl/seq.vhd
thomascobb: I'll change it

modules/seq/hdl/seq.vhd
thomascobb: `reset_i` was always ``0`` before, so was doing nothing. I have no idea whether the autogeneration framework initializes read registers to zero or no. Do you know @glennchid ?

modules/seq/hdl/seq.vhd
thomascobb: I'll change it

modules/seq/hdl/seq.vhd
Araneidae: Oh, I see: `TABLE_LENGTH` is a generic input parameter, which for generality is left as `std_ulogic_vector`.  Pity we don't have the `_i`/`_o` naming convention on these!

modules/seq/hdl/seq.vhd
Araneidae: Oh blimey.  I just spotted where these three are declared.  They're declared as local symbols, but they've go the port naming convention.  I think I'm opening a can of worms here...

Regarding initial values, no, their initial values are currently undefined.  Fortunately this isn't a change.

modules/seq/hdl/seq.vhd
glennchid: If it stays an slv I think you can use:
`if TABLE_LENGTH = (TABLE_LENGTH`range => '0') then`
to check for zeros, however I am never sure of synthesis support for this. What Tom had feels much safer!

glennchid: okay I can't figure out how to escape backticks in code comments: should read TABLE_LENGTH'range

Araneidae: Yes, but you can replace the rhs of the comparison with a simple `0`, that was my main point here.

modules/seq/hdl/seq.vhd
glennchid: Initial values for registers in SRAM based FPGAs will always default to '0' unless explicity set to '1', despite the fact that a simulator will consider them undefined.

Araneidae: Hmm.  I wasn't aware of this, I thought it was at the tool's discretion.  Makes sense, but certainly something I try to never rely on, seems like terrible practice!  Anyway, not relevant to this conversation, will mark it as resolved.

PR #43, opened 2021-03-11
Fix style issues

thomascobb:


modules/seq/hdl/seq.vhd
```diff
@@ -305,22 +305,22 @@ if rising_edge(clk_i) then
...
+                    TABLE_LINE_o <= (others => '0');
+                    LINE_REPEAT_o <= (others => '0');
                 elsif enable_rise = '1' then
                     -- rising ENABLE and trigger not met
-                    if (next_trig_valid  = '0') then
+                    if next_trig_valid  = '0' then
                         seq_sm <= WAIT_TRIGGER;
                     -- rising ENABLE and trigger met phase1
-                    elsif (next_frame.time1 /= to_unsigned(0,32)) then
+                    elsif next_frame.time1 /= to_unsigned(0,32) then
```
Araneidae, line 316: Also can write `0` instead of `to_unsigned(0,32)` here.

modules/seq/hdl/seq.vhd
```diff
@@ -305,22 +305,22 @@ if rising_edge(clk_i) then
...
                         next_ts <= next_frame.time1;
                         out_val <= next_frame.out1;
                         seq_sm <= PHASE_1;
                     -- rising ENABLE and trigger met and no phase 1
                     else
-                        if (next_frame.time2 /= to_unsigned(0,32)) then
+                        if next_frame.time2 /= to_unsigned(0,32) then
```
Araneidae, line 322: Same as above

modules/seq/hdl/seq.vhd
```diff
@@ -346,16 +346,16 @@ if rising_edge(clk_i) then
             -- State 2
             when WAIT_TRIGGER =>
                 -- trigger met
-                if (current_trig_valid = '1') then
+                if current_trig_valid = '1' then
                     -- trigger met
-                    if (current_frame.time1 /= to_unsigned(0,32)) then
+                    if current_frame.time1 /= to_unsigned(0,32) then
```
Araneidae, line 350: And so on ... won't note any more.

Araneidae (commented):
Why do we have local signals 
https://github.com/PandABlocks/PandABlocks-FPGA/blob/e2283fa782c445fc2c6ba87d288892d2655cc81c/modules/seq/hdl/seq.vhd#L99-L101 with the `_o` port naming convention?  This should be fixed.

It's also unfortunate that we have ports https://github.com/PandABlocks/PandABlocks-FPGA/blob/e2283fa782c445fc2c6ba87d288892d2655cc81c/modules/seq/hdl/seq.vhd#L51-L63 that _don't_ follow the naming convention, but I guess that's baked into the framework and lots of other code, so would be a much bigger job to fix.

thomascobb:
> Why do we have local signals with the `_o` port naming convention? This should be fixed.
>
> It's also unfortunate that we have ports that _don't_ follow the naming convention, but I guess that's baked into the framework and lots of other code, so would be a much bigger job to fix.

Yes, I'm not going to touch anything outside this file, we can schedule a bigger cleanup another time.

The convention that seems to be used here is that all ports are std_logic or std_logic_vector named:
- `some_input_i`
- `SOME_PARAMETER : `
- `some_output_o`
I'll uppercase the parameter so at least we're consistent, but what should we name the internal unsigned variables that will be converted on output? `SOME_PARAMETER_U`?

Araneidae:
I tend to use `name_in` for names that are assigned from `_i` ports, and `name_out` for names that are going to be assigned to `_o` ports.  This isn't a hard convention, just a convenient way of choosing sensible names.

thomascobb:
Ok, I went with TABLE_LINE for the read parameter, and TABLE_LINE_OUT for the unsigned int version of it

Araneidae:
I presume this all simulates and synthesises ok?

thomascobb:
Yes, and I've just tested it on a real PandA and it works...

Issue #41, opened 2021-02-22
Block SEQ: Errors in the sequencer block or in application?


jbeca:
### oscilloscope : show errors in sequence.
![scope](https://user-images.githubusercontent.com/9820764/111285303-4b583400-8641-11eb-86d0-8e38c97effe6.PNG)

### descripition:
The enable input Blocks are controled by SRGATE1.

The SRGATE 2 and 3 are controlled by the sequencer.

The connections are:
    SRGATE2.RST=SEQ1.OUTB
    SRGATE2.SET=SEQ1.OUTA
    SRGATE3.RST=SEQ1.OUTD
    SRGATE3.SET=SEQ1.OUTC

The inputs used for the sequencer are:
- a pos input from an encoder (SEQ1.POSA = INENC1.VAL)
- and bit from a rsing edge event come from a ttl input counter
      SEQ1.BITA = LUT1.OUT
      LUT1.TYPEA = Pulse-On-Rising-Edge
      LUT1.INPA=COUNTER8.CARRY
      COUNTER8.TRIG=TTLIN1.VAL

The sequencer Table:
Table line 1,2,3,4 are the sequence base.
This sequence are repeted with the table lines (5,6,7,8) and (9,10,11,12).

**Erreurs** are in sequence ligne 5,8,9,12.  (**with regard to the reference sequence, there is or not a signal)**
(There is no signal generated by SEQ to control the srgate2.)
(the expected signal is indicated in gray)

**Erreurs** : The table line 13 is a test "do nothing" but the outputs change.
(SEQ does not generate signals to control the srgate, but the srgate2 ant 3 change state)

### panda parameters files:
[program.txt](https://github.com/PandABlocks/PandABlocks-FPGA/files/6023970/program.txt)
[table.txt](https://github.com/PandABlocks/PandABlocks-FPGA/files/6023778/table.txt)

jbeca:
### oscilloscope : show errors in sequence.
![image](https://user-images.githubusercontent.com/9820764/108745222-20dff300-753b-11eb-9f31-a1b99214e93a.png)

### sequencer table :
[table.txt](https://github.com/PandABlocks/PandABlocks-FPGA/files/6023778/table.txt)

shu-soleil:
Jerome, could you try to set the SEQ1.TABLE with phase time > 1 (8ns) on each line ?

thomascobb:
I've reproduced this, I'll look into it...

[...]

PR #45, opened 2021-03-24
Fix sequencer to work with 1 clock tick PHASE2

thomascobb:
`load_next` was introducing a 1 clock delay on getting the next line, stopping it working for a 1 clock tick PHASE2. The following clause was inserted in the past to activate it in PHASE1, but the logic was wrong, causing #41 
`(current_frame = next_frame and seq_sm = PHASE_1 and unsigned(PRESCALE) < to_unsigned(2,32) and (current_frame.time2 = 0 or current_frame.time2 = to_unsigned(1,32)))`

This PR moves the selection of `seq_raddr` into combinatorial logic backed by a clock process that is always updating `seq_raddr_next`

All tests passing

### DIFF
--- a/modules/seq/hdl/seq.vhd
+++ b/modules/seq/hdl/seq.vhd
@@ -134,8 +134,6 @@ signal table_ready_dly      : std_logic;
 
 signal next_ts              : unsigned(31 downto 0);
 
-signal test_load_next       : std_logic;
-
 begin
 
 -- Block inputs.
@@ -182,7 +180,7 @@ port map (
     TABLE_START         => TABLE_START_WSTB,
     TABLE_DATA          => TABLE_DATA,
     TABLE_WSTB          => TABLE_DATA_WSTB,
-    TABLE_LENGTH        => TABLE_FRAMES,
+    TABLE_FRAMES        => TABLE_FRAMES,
     TABLE_LENGTH_WSTB   => TABLE_LENGTH_WSTB
 );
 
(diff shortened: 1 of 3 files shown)


modules/seq/hdl/seq.vhd
```diff
@@ -273,11 +271,13 @@ STATE(31 downto 3) <= (others => '0');
...
-                      (current_frame = next_frame and seq_sm = PHASE_1 and unsigned(PRESCALE) < to_unsigned(2,32)
-                      and (current_frame.time2 = 0 or current_frame.time2 = to_unsigned(1,32)))
-                      else '0';
+-- combinatorial logic so we can load a new row in 1 clock tick
+-- commented out load_next in SEQ_FSM shows where this logically sits
+load_next <= '1' when (seq_sm = WAIT_ENABLE and enable_rise = '1') or (
+    seq_sm = PHASE_2 and presc_ce = '1' and tframe_counter = next_ts -1
+    and last_line_repeat = '1') else '0';
+reset_table <= '1' when enable_i = '0' or TABLE_START_WSTB = '1' else '0';
+
```
Araneidae, line 280: I see, this explains the commented out `reset_table` assignments below.  Still think that's a bad idea, it looks too much like a mistake.

This is seriously scary looking code!  Might be worth adding the following function to the general support library:
```vhdl
function to_std_ulogic(bool : boolean) return std_ulogic is
begin
    if bool then
        return '1';
    else
        return '0';
    end if;
end;
```
This would at least let you write
```vhdl
load_next <= to_std_ulogic(
    (seq_sm = WAIT_ENABLE and enable_rise = '1') or
    (seq_sm = PHASE_2 and presc_ce = '1' and tframe_counter = next_ts -1
         and last_line_repeat = '1'));
```
This might be a candidate for the otherwise hideous with/select statement:
```vhdl
with seq_sm select load_next <=
    enable_rise when WAIT_ENABLE,
    presc_ce and to_std_ulogic(tframe_counter = next_ts-1) and last_line_repeat
        when PHASE_2,
   '0' when others;
```
Or maybe a process statement (I do so despise VHDL's indecision about statement syntax):
```vhdl
process (all) begin
    case seq_sm is
        when WAIT_ENABLE =>
            load_next <= enable_rise;
        when PHASE_2 =>
            if tframe_counter = next_ts - 1 then
                load_next <= presc_ce and last_line_repeat;
            else
                load_next <= '0';
            end if;
        when others =>
            load_next <= '0';
    end case;
end process;
```
Sorry to be so indecisive here.  This whole entity is pretty horrifying, particularly the gigantic state machine below.

Araneidae (approved):
Nothing hugely useful to say here.  I've not looked at the logic of your change, but I see that you pulled some logic into unregistered logic to bring it forward a tick.  Think my style comments aren't hugely useful this time, there probably is no clear way to write what's going on here.

modules/seq/hdl/seq.vhd
```diff
@@ -297,7 +297,7 @@ if rising_edge(clk_i) then
         out_val <= (others => '0');
         active <= '0';
         if seq_sm /= LOAD_TABLE then
-            reset_table <= '1';
+--          reset_table <= '1';
```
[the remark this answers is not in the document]
PaulJG: Agree

modules/seq/hdl/seq.vhd
PaulJG: I don't see how the examples above would simplify the code.
Yes the state machine is big that why good documentation is also important.

PaulJG:
Looks OK to me

modules/seq/hdl/seq.vhd
Araneidae: It's not so much about "simplifying" the code as making the intent of the code clearer, and switching on `seq_sm` does at least have the virtue of making it clear that the behaviour of `load_next` really is driven by the state machine.  I showed all the forms above because none of them work terribly well for me!

PR #47, opened 2021-05-21
Adapt FPGA loading to use the FPGA manager

hir12111:
This is required to use the latest kernel version and should be merged at the same time that this [PR](https://github.com/PandABlocks/PandABlocks-rootfs/pull/18)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -195,6 +195,7 @@ hdl_timing: $(TIMING_BUILD_DIRS)
 # FPGA build
 
 FPGA_FILE = $(FPGA_BUILD_DIR)/panda_top.bit
+FPGA_BIN_FILE = $(FPGA_BUILD_DIR)/panda_top.bin
 SLOW_FPGA_FILE = $(SLOW_FPGA_BUILD_DIR)/slow_top.bin
 
 FPGA_DEPENDS =
@@ -212,6 +213,13 @@ else
             TOP=$(TOP) TARGET_DIR=$(TARGET_DIR) BUILD_DIR=$(dir $@) \
             IP_DIR=$(IP_DIR)
 endif
+
+$(FPGA_BIN_FILE): $(FPGA_FILE)
+	cd $(FPGA_BUILD_DIR) && \
+        echo -e "all:\n{\n    $(FPGA_FILE)\n}\n" > bs.bif && \
+        source $(VIVADO) && \
+        bootgen -image bs.bif -arch zynq -process_bitstream bin && \
+        mv $(FPGA_FILE).bin $@
 
 $(SLOW_FPGA_FILE): $(AUTOGEN_BUILD_DIR) $(SLOW_FPGA_DEPENDS)
 	mkdir -p $(dir $@)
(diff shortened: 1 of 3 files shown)


CONFIG.example
```diff
@@ -16,6 +16,9 @@ LM_LICENSE_FILE = 2100@diamcslicserv01.dc.diamond.ac.uk
 PANDA_ROOTFS = /home/mga83/targetOS/PandABlocks-rootfs
 # MAKE_ZPKG = $(PANDA_ROOTFS)/make-zpkg
 
+# Location of SDK used to find bootgen.
+SDK_ROOT = /dls_sw/apps/FPGA/Xilinx/SDK/2016.4
+
```
glennchid, line 21: why is this pointing to Vivado 2016.4?

Makefile
```diff
@@ -25,6 +26,9 @@ DEFAULT_TARGETS = zpkg
 # and editing as appropriate.
 include CONFIG
 
+ifdef SDK_ROOT
+    BOOTGEN = $(SDK_ROOT)/bin/bootgen
+endif
```
glennchid, line 31: we can just call bootgen directly by sourcing the vivado shell script for whatever version we are using

Makefile
```diff
@@ -213,6 +218,11 @@ else
             IP_DIR=$(IP_DIR)
 endif
 
+$(FPGA_BIN_FILE): $(FPGA_FILE)
+	cd $(FPGA_BUILD_DIR) && \
+        echo -e "all:\n{\n    $(FPGA_FILE)\n}\n" > bs.bif && \
+        $(BOOTGEN) -image bs.bif -arch zynq -process_bitstream bin
+
```
glennchid, line 225: Is there anything special about the format of the bin file? Presumably it is just the bitstream with the header removed. We can tell bitgen to write out the bin file instead of/as well as the bit file.

glennchid (changes requested):
I think we can do this in a much simpler way without the changes to the Makefile and CONFIG. Just replacing the .bit for .bin, if you really need a bin file. Does supplying a bit file directly not work?

CONFIG.example
hir12111: SDK_ROOT is only used for running bootgen, I suppose there is a better way.
I pointed to the same version that Panda rootfs is using so that both uses same bootgen

Makefile
hir12111: Thanks, I will investigate that possibility

Makefile
hir12111: The only thing special about the bin file is that it is required by the FPGA manager

CONFIG.example
hir12111: This was removed

glennchid (approved):
Agreed that we should merge this commit as-is alongside changes to the PandABlocks-rootfs repo, but look to revert the bin file generation in Makefile if FPGA manager can work with the bit or bin file produced by bitgen in later versions of Vivado.

PR #53, opened 2021-08-27
Moving slowFPGA build to separate repo

glennchid:
To make it easier to build the FPGA zpkg with CI actions in github, we decided to try separating the slow-fpga from the Zynq build. Recent versions of Vivado no longer require a licence file or server for webpack parts, whereas Xilinx ISE does. The slow-fpga will be built locally, and as it is changed only infrequently, CI is unnecessary for this. 

Please review the following changes in association with the new repo for the Slow FPGA:
https://github.com/PandABlocks/PandABlocks-slowFPGA

There are now several source files that are now duplicated in both repos, mainly support modules found in common/hdl.
'slow_defines.vhd' must be kept in both repos and in sync, as it contains the register mapping for the slow-fpga registers.


thomascobb (approved):

CONFIG.example
```diff
@@ -12,8 +12,6 @@
 
 # Definitions needed for FPGA build
 export VIVADO = /dls_sw/FPGA/Xilinx/Vivado/$(VIVADO_VER)/settings64.sh
-export ISE = /dls_sw/FPGA/Xilinx/14.7/ISE_DS/settings64.sh
-export LM_LICENSE_FILE = 2100@diamcslicserv01.dc.diamond.ac.uk
```
Araneidae, line 16: Why have these been removed?  The `LM_LICENSE_FILE` is definitely still needed...

modules/system/hdl/system_registers.vhd
```diff
@@ -39,6 +39,32 @@ end system_registers;
...
+-- TTLIN TERM Address List
+constant TTLTERM_ADDR_LIST  : page_array(TTLIN_NUM-1 downto 0) := (
+                                TO_SVECTOR(TTLIN6_TERM, PAGE_AW),
+                                TO_SVECTOR(TTLIN5_TERM, PAGE_AW),
+                                TO_SVECTOR(TTLIN4_TERM, PAGE_AW),
+                                TO_SVECTOR(TTLIN3_TERM, PAGE_AW),
+                                TO_SVECTOR(TTLIN2_TERM, PAGE_AW),
+                                TO_SVECTOR(TTLIN1_TERM, PAGE_AW)
+                            );
+
```
Araneidae, line 67: ~What has this got to do with moving the slowFPGA build out?~

Oh, I see.  Tricky ... this list of definitions is also needed by the slow FPGA, as it's very architecture specific and is managed by the slow FPGA.  Hmmm.

targets/PandABox/hdl/slow_defines.vhd
```diff
@@ -54,35 +46,10 @@ constant FMC_15VP           : natural := 15;
...
--- TTLIN TERM Address List
-constant TTLTERM_ADDR_LIST  : page_array(TTLIN_NUM-1 downto 0) := (
-                                TO_SVECTOR(TTLIN6_TERM, PAGE_AW),
-                                TO_SVECTOR(TTLIN5_TERM, PAGE_AW),
-                                TO_SVECTOR(TTLIN4_TERM, PAGE_AW),
-                                TO_SVECTOR(TTLIN3_TERM, PAGE_AW),
-                                TO_SVECTOR(TTLIN2_TERM, PAGE_AW),
-                                TO_SVECTOR(TTLIN1_TERM, PAGE_AW)
-                            );
-
```
Araneidae, line 82: No, I don't get it.  Why have these moved?  These are indeed specific to the slow FPGA because it's in control of these items.

Araneidae (commented):
A few questions.  I think one of the changes to `CONFIG.example` doesn't belong here, and I'm a bit puzzled by the move of a chunk of definitions from `targets/PandABox/hdl/slow_defines.vhd` to `modules/system/hdl/system_registers.vhd`.

Other than that looks fine.  <sub>I do think changing `bits` to `bit` for the bit-file Makefile identifier is mostly churn, as a bitfile contains bits... but that's not really worth a review comment.</sub>

targets/PandABox/hdl/slow_defines.vhd
glennchid: As mentioned in the PR above, 'slow_defines.vhd' is needed by both FPGA repos and needs to be kept in sync between the two. These three constants used to be defined in 'slow_defines' but are used only by the Zynq FPGA, and refer to other constants which are not used or known about by the Slow FPGA. As these three constants are only used in 'system_registers.vhd', the obvious solution was to move them to there.

modules/system/hdl/system_registers.vhd
glennchid: See below

CONFIG.example
glennchid: Xilinx ISE is no longer required by this repo.

LM_LICENCE_FILE -  I am not so bothered about this. It is actually no longer needed for the regular PandA builds using recent versions of Vivado, as webpack devices no longer require a licence file. If targeting non-webpack devices or using licenced IP, the licence server can be set in the personal CONFIG file (I have this set in my bash profile).  As CI builds in the cloud do not need to know about our local licence server, I would prefer to have no mention of licencing in the public repos at all.

Here is suggestion though, if you prefer:

`# Specify path to licence server, if required`
`# export LM_LICENCE_FILE = `

Araneidae: I think it probably is worth adding just the empty `LM_LICENCE_FILE` then for those applications which will need a licence.

Araneidae (approved):
Looks ok

Issue #56, opened 2021-09-20
PCAP "Driver data overrun" error


shu-soleil:
First issue was posted in PandABlocks-server project: "https://github.com/PandABlocks/PandABlocks-server/issues/16": Under some configurations, pcap acquisitions always finished with the "Driver data overrun" error. It's reproducble on different firmware versions and hardware platforms. Sometimes when this error produced, the log file "/var/log/messages" showed also an unexpected IRQ received by the driver.

PR #57, opened 2021-09-20
pcap: fix driver overrun issue caused by the last 2 IRQs too close

shu-soleil:
to fix issue #56: "Driver data overrun" error occured when the last DMA IRQ is too close to the completion IRQ.

### DIFF
--- a/modules/pcap/hdl/pcap_dma.vhd
+++ b/modules/pcap/hdl/pcap_dma.vhd
@@ -320,7 +320,8 @@ if rising_edge(clk_i) then
                 -- Position compare completed
                 elsif (pcap_completed = '1' and writing_sample = '0') then
                     last_tlp <= '1';
-                    if (fifo_count = 0) then
+                    -- if last_tlp, wait for next buffer address assigned by PS
+                    if (fifo_count = 0 and next_dmaaddr_valid = '1') then
                         pcap_fsm <= IRQ;
                     else
                         dma_start <= '1';


modules/pcap/hdl/pcap_dma.vhd
```diff
@@ -320,7 +320,8 @@ if rising_edge(clk_i) then
                 -- Position compare completed
                 elsif (pcap_completed = '1' and writing_sample = '0') then
                     last_tlp <= '1';
-                    if (fifo_count = 0) then
+                    -- if last_tlp, wait for next buffer address assigned by PS
+                    if (fifo_count = 0 and next_dmaaddr_valid = '1') then
```
Araneidae, line 324: The logic here looks pretty well identical to the timeout handling block above: https://github.com/PandABlocks/PandABlocks-FPGA/blob/993fb7b684dfd46698b6222ebd12413723c95aa0/modules/pcap/hdl/pcap_dma.vhd#L306 Does this line also need to be qualified by `next_dmaaddr_valid`?

Araneidae (commented):
A simple fix, definitely worth applying.

Having found a bug in this code, it might be helpful to review the design in a bit more depth.  I'm not at all sure I understand what's going on here in enough detail.  It would good to have a narrative comment that explains the logic, but don't think that's part of this pull request.

One thing that strikes me as slightly odd: apart from the `pcap_error` test here https://github.com/PandABlocks/PandABlocks-FPGA/blob/993fb7b684dfd46698b6222ebd12413723c95aa0/modules/pcap/hdl/pcap_dma.vhd#L300 all the logic for transitioning out of `pcap_fsm = ACTV` state is qualified by `writing_sample = '0'`, which is essentially a two tick window around `pcap_wstb_i`.  Any idea why this is done?

modules/pcap/hdl/pcap_dma.vhd
shu-soleil: The current logic is: the timeout will active an interrupt IRQ where it checks the value of next_dmaaddr_valid (equals to the driver acknowledge of last IRQ), if it's not done then an error of "Driver data overrun" will be raised to show that data are arrived too fast for the PS. 
We can either keep this logic, or wait for next_dmaaddr_valid=1 before sending any IRQ - in which case the "Driver overrun error" will never be raised and can be removed. 

 https://github.com/PandABlocks/PandABlocks-FPGA/blob/993fb7b684dfd46698b6222ebd12413723c95aa0/modules/pcap/hdl/pcap_dma.vhd#L369-L370

> The logic here looks pretty well identical to the timeout handling block above:
> 
> https://github.com/PandABlocks/PandABlocks-FPGA/blob/993fb7b684dfd46698b6222ebd12413723c95aa0/modules/pcap/hdl/pcap_dma.vhd#L306
> 
> Does this line also need to be qualified by `next_dmaaddr_valid`?

shu-soleil: No idea for this question ...
> all the logic for transitioning out of pcap_fsm = ACTV state is qualified by writing_sample = '0', which is essentially a two tick window around pcap_wstb_i. Any idea why this is done?

Araneidae: Ah, I see.  No, it certainly would make sense to raise the "Driver overrun error" here.

I'm still a little concerned: there is a one tick delay between making the decision to transition `pcap_fsm` to the `IRQ` state and latching the interrupt register.  Potentially this still leaves room for the incorrect interrupt reason being reported.  Unfortunately, given the way the code is written, we need to look at the life-cycle of every flag to check this; having done this, I _think_ we're ok, but I'm not sure.

In passing, I note that `irq_flags(0)` (DMA active) is not set.  Guess I should raise a separate issue for this.

Araneidae:
I am happy with this pull request and would like it to be merged, but would feel happier if we understood the code better!

modules/pcap/hdl/pcap_dma.vhd
shu-soleil: "writing_sample" going with two ticks around "pcap_wstb_i" might for synchronization with both delayed signal (eg. "pcap_completed" one tick delayed to "pcap_done_i") and non delayed signals like "pcap_timeout" ? 

If we assign pcap_completed to (pcap_completed and pcap_done_i), maybe writing_sample could be simply <= pcap_wstb_i.

Araneidae:
@thomascobb, I suggest we merge this right away, but a little more love and care to PCAP would not go amiss.  I notice the CI check has been marked as "Some checks haven't completed yet" since yesterday afternoon; is there something snagging here?

thomascobb:
The CI PR hasn't merged yet, so is still using the broken travis implementation. There's a couple of things left to do on it, so I'll just merge this one now

PR #59, opened 2021-09-21
pcap: avoid pcap_fsm jumping in ACTV/DMA states before last IRQ

shu-soleil:
Sorry to add a patch to the last PR #57, I found with ILA that the last change will make pcap_fsm jumping in ACTV-DMA-IS_FINISHED states before achieving the last IRQ state, as shown in this image: 
ILA_Overrun_200-00024ms_3pts_2 1-14-gbe2a2ea

So I fixed it by this new PR.


thomascobb:
Need to discuss this at the next collaboration meeting

PR #67, opened 2021-11-04
Add support for SoC Mercury XU5 on base board ST1

lyodaom:
This was successfully tested on a ME-XU5-4EV-1I-D11E-G1 mounted on a ST1.


glennchid (approved):

thomascobb (commented):
I don't understand half of this PR, but as long as @glennchid understands it then the bits I know about look fine

PR #69, opened 2021-11-09
Add new target NamcZynq board based on a SoC Zynq 7045

shu-soleil:


Araneidae:
What is the status of this PR?

PR #70, opened 2022-03-24
Building dtc from source and removing binary blob

glennchid:
* previous blob was an old 32-bit version requiring 32-bit libraries


common/fpga.make
```diff
@@ -228,14 +232,18 @@ u-boot-src: $(U_BOOT_SRC)
 .PHONY: u-boot-src
 # -----------------------------------------------------------------------------------
 
-$(DEVTREE_DTB): $(SDK_EXPORT) $(TARGET_DTS)
+$(DEVTREE_DTB): $(SDK_EXPORT) $(TARGET_DTS) | $(DEVTREE_DTC)
```
coretl, line 235: What does the `|` mean?

glennchid: Order-only pre-requisite - checks it exists but does not trigger rebuild of DTB if DTC timestamp gets changed. I think this is appropriate as we are not changing the dtc sources, but just need to binary to exist.

coretl: In which case it looks fine

common/fpga.make
```diff
@@ -228,14 +232,18 @@ u-boot-src: $(U_BOOT_SRC)
...
 	cp $(TARGET_DTS) $(DEVTREE_DTS)/
 	sed -i '/dts-v1/d' $(DEVTREE_DTS)/system-top.dts
 	gcc -I dts -E -nostdinc -undef -D__DTS__ -x assembler-with-cpp \
 	  -o $(DEVTREE_DTS)/system-top.dts.tmp $(DEVTREE_DTS)/system-top.dts
 	@echo "Building DEVICE TREE blob ..."
 	$(DEVTREE_DTC) -f -I dts -O dtb -o $@ $(DEVTREE_DTS)/$(notdir $(TARGET_DTS))
 
+$(DEVTREE_DTC): | $(SRC_ROOT)
```
lyodaom, line 243: that dependency doesn't look right, I suggest splitting this target in 2 (similar way to the other targets in the same file):
```Makefile
DTC_SRC = ....

$(DTC_SRC): | $(SRC_ROOT)
	$(call EXTRACT_FILE,dtc-$(DTC_TAG).tar.gz,$(MD5_SUM_dtc-$(DTC_TAG)))
	chmod -R a-w $(DTC_SRC)

$(DEVTREE_DTC): $(DTC_SRC)
	$(MAKE) -C $(SRC_ROOT)/dtc-$(DTC_TAG) NO_PYTHON=1
```

common/fpga.make
lyodaom: I would suggest not using the `|` for this case, I agree that changing DTC sources is not planned (in which case, the timestamp check doesn't matter), but if we ever have a reason to do it in future, this will get in our way for testing.

common/fpga.make
glennchid: done

PR #73, opened 2022-04-01
Allow passing fpga options in the application file

lyodaom:
It also allows disabling options activated by target file if the option is prefixed by ! (exclamation mark)

A basic unit tests was added.


Araneidae (approved):
Looks ok to me.

PR #76, opened 2022-05-24
Finedelay

lyodaom:
This change provides fine delay control in TTLOUT and LVDSOUT modules. It requires 2 parameters:
QUARTER_DELAY: number of 1/4 tick (range 0-3)
FINE_DELAY: counter used in ODELAY (range 0-31)

The ODELAY counter is supposed to provide 78 ps resolution ~~, however, some tests using a scope shows around twice that value~~.
Another thing to consider is that the ODELAY is not available in all I/O banks, so, there is a possibility that other targets will not benefit from it.

### DIFF
--- a/common/hdl/mmcm_clkmux.vhd
+++ b/common/hdl/mmcm_clkmux.vhd
@@ -12,8 +12,10 @@ port (fclk_clk0_ps_i      : in  std_logic;
       clk_sel_i           : in  std_logic_vector(1 downto 0);
       linkup_i            : in  std_logic;
       sma_pll_locked_o    : out std_logic;
+      fclk_locked_o       : out std_logic;
       clk_sel_stat_o      : out std_logic_vector(1 downto 0);
-      fclk_clk0_o         : out std_logic
+      fclk_clk0_o         : out std_logic;
+      fclk_clk0_2x_o      : out std_logic
       );
 
 end mmcm_clkmux;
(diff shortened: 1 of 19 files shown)


targets/PandABox/hdl/PandABox_top.vhd
```diff
@@ -312,6 +313,12 @@ port map (
         IB              =>      GTXCLK1_N
     );
 
+idelayctrl_inst : IDELAYCTRL port map (
+    REFCLK => FCLK_CLK1_PS,
+    RST => '0',
```
Araneidae, line 316: Might be worth providing a reset here, don't think it has to be synchronous with `REFCLK`.

common/hdl/mmcm_clkmux.vhd
```diff
@@ -43,27 +49,20 @@ begin
...
-                sma_pll_reset <= '1';
-            end if;
-        end if;
-    end if;
-end process ps_sma_reset_pll;
+pll_autoreset_inst1 : entity work.pll_autoreset port map (
+    clk_i => ps_fclk_bufh,
+    pll_locked_i => sma_pll_locked,
+    pll_reset_o => sma_pll_reset
+);
```
Araneidae, line 56: I approve of this new entity ... but it's puzzling to me to see the commit that introduced it so separated from the commit that uses it!

targets/PandABox/const/PandABox-clks_impl.xdc
```diff
@@ -3,10 +3,24 @@
 # -------------------------------------------------------------------
 
 set_clock_groups -asynchronous -group clk_fpga_0
+set_clock_groups -asynchronous -group clk_fpga_1
 set_clock_groups -asynchronous -group EXTCLK_P
 set_clock_groups -asynchronous -group GTXCLK0_P
 set_clock_groups -asynchronous -group GTXCLK1_P
```
Araneidae, line 10: This complex set of cross-clock exclusions is sufficiently obscure that I think it needs a comment to explain what's going on here!

Araneidae (approved):
One general comment: it's a bit strange to have so many tiny commits introducing changes to one file at a time; I'd prefer to see the commits structured around the change being made.  From this view, I think I see three main themes:

- Introducing the new clocks, including the rather horrible cross-clock constraints.  Maybe one or two commits for this work?
- Creating the new entity.
- Implementing it.

Araneidae:
Hmm.  Looks like we have some timing constraint failures in other builds.  Fear that this needs looking at...

common/hdl/mmcm_clkmux.vhd
lyodaom: Done

targets/PandABox/hdl/PandABox_top.vhd
lyodaom: Done

lyodaom:
The formula for the ODELAY resolution is 1/(32*2*Fref), we can only have a reference clock around 200 MHz for speed grade -1 (which I think it's the one in the picozed), therefore we are supposed to get a resolution around 78 ps.

Issue #51, opened 2021-08-03
Add support for calculated registers


thomascobb:
As per https://github.com/PandABlocks/PandABlocks-server/issues/9

tomtrafford:
I was just having a quick look at how to implement this. I've made an example to how I understand it to work. For a block.ini such as:
```
[READ1]
type: read
description:
extension: test_extension1

[READ2]
type: read
description:
extension: test_extension1

[WRITE1]
type: write
description:
extension: test_extension2

[WRITE2]
type: write
description: 
extension: test_extension2

[READ3]
type: read
description:
extension: test_extension3

[READ4]
type: read
description:
extension: test_extension3

[WRITE3]
type: write
description:
extension: test_extension3

[WRITE4]
type: write
description: 
extension: test_extension3
```
A generated registers file would look like:

```
    READ1 READ2                    X test_extension1
    W WRITE1 WRITE2                X test_extension2
    READ3 READ4 W WRITE3 WRITE4    X test_extension3
```

So if multiple fields have the same extensions defined they will combine the entries in the Registers file depending on the field type. Have I understood this correctly?

Araneidae:
Hi @tomtrafford, this isn't right.  Have a look at https://github.com/PandABlocks/PandABlocks-server/blob/c8faaa58459743a975d363607d2979dbaf12108d/python/sim_config/registers#L246-L255 for an example register file.  The corresponding config file doesn't show the extensions at all.

The syntax of the extension register definition is described here: https://github.com/PandABlocks/PandABlocks-server/blob/c8faaa58459743a975d363607d2979dbaf12108d/server/extension.c#L238-L246:

Read registers:

    [register]* X extension-name

Write registers:

    [register]* [W [register]*] X extension-name

The meaning of this is: the first named registers will be read and passed to the extension server as arguments to the function `extension-name`.  If this is a write register then this function will be expected to return a tuple of values which will then be written to the registers after the optional `W`.

Hope this makes more sense.

Araneidae:
Hmm.  Why can't I link to lines of code anymore?

thomascobb:
@Araneidae I can't get my head around this.

Let's consider a simple example.

I want to have 2 write records, A and B, and 2 read records AVERAGE and DIFFERENCE. Whenever A or B is written to, the extension should calculate AVERAGE and DIFFERENCE:

```
    def write(self, block, a, b):
        average = int((a+b)/2)
        difference = a-b
[...]

PR #79, opened 2022-06-29
Added support for write extensions

Araneidae:
This needs to be completed and verified!


tomtrafford:
I have added some tests to the python tests which show the autogenerated config, description and registers for a block.ini which uses both write and read registers. This should match what was discussed in #51. @Araneidae I don't seem to be able to add you as reviewer.

Araneidae:
> @Araneidae I don't seem to be able to add you as reviewer.

That's strange, nor do I!

Maybe that's because I create the initial pull request?  That somewhat sucks if so.

Araneidae:
I have two questions to start with:
1. Have the existing extension implementations been updated to work with the new extension framework?
2. Are you using the new `ExtensionHelper` class as recommended [here](https://pandablocks.github.io/PandABlocks-server/master/extension.html)?
I forget which extensions already exist and how they are affected by the new server.

Araneidae:
A quick search identifies:
```
./modules/fmc_acq427/extensions/fmc_acq427.py
./modules/system/extensions/system.py
./modules/us_system/extensions/us_system.py
```
and it looks as if these have all already been updated to create an `Extension` class of the correct format.

I was probably asking the wrong question then, sorry about that.

tomtrafford:
@Araneidae Can you let me know if you think anything further is required for this Pull request? I have made python tests to show that the block.ini files creates the Config and [Registers](tests/python/test_data_calc_extensions/app-expected/config_d/registers) file for interval and dummy matching the[ sim config](https://github.com/PandABlocks/PandABlocks-server/blob/master/python/sim_config/registers) from the server, as well as for the fmc_ acq_427 as we found that was a special case.

common/python/configs.py
```diff
@@ -200,6 +219,10 @@ def __init__(self, name, number=-1, prefix='', extension=''):
         self.prefix = prefix
         #: For an extension field, the register path
         self.extension = extension
+        #: If there is a write extension
+        self.write_extension = write_extension
+        #: If there is a write extension
```
EmilioPeJu, line 224: you probably mean `read`

common/python/configs.py
```diff
@@ -241,6 +264,10 @@ def __init__(self, name, number, type, description, extra_config):
         #: Store the extension register info
         self.extension = extra_config.pop("extension", None)
         self.extension_reg = extra_config.pop("extension_reg", None)
+        self.extension_write = extra_config.pop("extension_write", "")
+        self.extension_read = extra_config.pop("extension_read", "")
+        self.extension_nums = []
+        self.no_config = 0
```
EmilioPeJu, line 270: is there a reason why this couldn't be `bool` instead? I think it would be easier to read

common/python/configs.py
```diff
@@ -465,8 +505,17 @@ def register_addresses(self, counters):
...
 
+class CalcExtensionFieldConfig(ParamFieldConfig):
+    """These fields act in the same way as write record from the VHDL generation 
+    point of view, but do not have a config entry"""
+    type_regex = "(extension_write|extension_read)"
+
+    def register_addresses(self, counters):
+        # type: (FieldCounter) -> None
+        super(CalcExtensionFieldConfig, self).register_addresses(counters)
+        self.no_config=1
```
EmilioPeJu, line 518: Missing spaces around `=` and I suggest using `True` instead

common/python/generate_app.py
```diff
@@ -53,13 +53,14 @@ def jinja_env(path):
...
-    def __init__(self, app, app_build_dir):
-        # type: (str, str) -> None
+    def __init__(self, app, app_build_dir, testPath=""):
+        # type: (str, str, str) -> None
         # Make sure the outputs directory doesn't already exist
         assert not os.path.exists(app_build_dir), \
             "Output dir %r already exists" % app_build_dir
         self.app_build_dir = app_build_dir
         self.app_name = app.split('/')[-1].split('.')[0]
+        self.testPath=testPath
```
EmilioPeJu, line 63: Missing spaces around `=`

common/python/generate_app.py
```diff
@@ -123,7 +124,13 @@ def parse_ini_files(self, app):
         self.process_fpga_options(
             ini_get(app_ini, '.', 'options', ''))
         # Implement the blocks for the soft blocks
-        self.implement_blocks(app_ini, "modules", "soft")
+        # If a test path has been given, use it for location of blocks.
+        # Otherwise blocks should be in moudles directory
+        if self.testPath:
+            path=self.testPath
+        else:
+            path="modules"
```
EmilioPeJu, line 132: ditto

common/python/generate_app.py
```diff
@@ -158,6 +165,7 @@ def implement_blocks(self, ini, path, type):
                 # for carrier block
                 block = BlockConfig(section, type, number, ini_path, siteNumber)
                 block.register_addresses(self.counters)
+                block.generateCalcExtensions()
```
EmilioPeJu, line 168: Could we use a consistent naming convention for functions? I think `generate_calc_extensions()` would be more pythonic

common/python/configs.py
```diff
@@ -263,10 +290,23 @@ def make_reg_name(r):
...
+                        result.extend(str(' '.join(str(num) )))
+                # If there are write extensions add W and then any register numbers
+                if r.write_extension:
+                    result.extend(['W'])
+                    for num, ext_type in self.extension_nums:
+                        if ext_type == "write":
+                            result.extend(str(' '.join(str(num) )))
                 result.extend(['X', r.extension])
+            else:
+                if r.number >= 0:
```
EmilioPeJu, line 308: `else: if ` can be replaced with `elif` in this case

common/python/configs.py
```diff
@@ -172,6 +174,23 @@ def generateInterfaceConstraints(self):
             if constraint not in self.interfaceConstraints:
                 self.interfaceConstraints.append(constraint)
 
+    def generateCalcExtensions(self):
+        # Iterate through the fields and add any with writeExtension type to the list
+        for field in self.filter_fields("extension_.*"):
+            extension = (field.name, field.registers[0].number)
+            self.extension=self.name.lower()
```
EmilioPeJu, line 181: More missing spaces, using `pycodestyle` to check against pep8 would be nice

EmilioPeJu:
Can I suggest that we simplify things for the time being?
I can see the server doesn't have any consideration for the `read_extensions` side and it's not needed in any use-case yet, I think we should support only the following cases(which the server seems to support):
extensions that provides 1 value, like the system block does
extensions that can write to a number of registers, like the interval example block

Therefore, I would remove the other mixed cases and the `read_extensions` part (which I don't understand why it generates write-only registers)

The register file would be then become simpler for extension related fields:
either
name X extension-specific-name

or
name W<register-offset-list> extension-specific-name

tests/python/test_data_calc_extensions/app-expected/hdl/dummy_ctrl.vhd
```diff
@@ -0,0 +1,114 @@
...
+            DUMMY_READ_0_wstb <= '0';
+            DUMMY_READ_1_wstb <= '0';
+            DUMMY_READ_2_wstb <= '0';
+            DUMMY_READ_3_wstb <= '0';
+            DUMMY_WRITE_wstb <= '0';
+            if (write_strobe_i = '1') then
+                -- Set the specific write strobe that has come in
+                case write_addr is
+                    when DUMMY_DUMMY_READ_0_addr =>
+                        DUMMY_READ_0 <= write_data_i;
```
EmilioPeJu, line 74: `DUMMY_READ_0` was declared as `in` parameter, as I suggested, I would remove everything to do with the `read_extensions` side of the change

Araneidae:
It's worth referencing back to the original specification, https://github.com/PandABlocks/PandABlocks-FPGA/issues/51#issuecomment-897704554 and https://github.com/PandABlocks/PandABlocks-server/issues/9#issuecomment-810045019.

@EmilioPeJu , are you proposing dropping all the read registers?  This will mean that extensions have no facility for reading the FPGA state if needed.

tests/python/test_data_calc_extensions/app-expected/hdl/dummy_ctrl.vhd
Araneidae: Yes, this code should be _reading_ from `DUMMY_READ_0` etc and only writing to `DUMMY_WRITE`.

tests/python/test_data_calc_extensions/app-expected/hdl/dummy_ctrl.vhd
```diff
@@ -0,0 +1,114 @@
...
+                        DUMMY_READ_1_wstb <= '1';
+                    when DUMMY_DUMMY_READ_2_addr =>
+                        DUMMY_READ_2 <= write_data_i;
+                        DUMMY_READ_2_wstb <= '1';
+                    when DUMMY_DUMMY_READ_3_addr =>
+                        DUMMY_READ_3 <= write_data_i;
+                        DUMMY_READ_3_wstb <= '1';
+                    when DUMMY_DUMMY_WRITE_addr =>
+                        DUMMY_WRITE <= write_data_i;
+                        DUMMY_WRITE_wstb <= '1';
```
Araneidae, line 87: This looks correct.

common/python/configs.py
```diff
@@ -241,6 +264,10 @@ def __init__(self, name, number, type, description, extra_config):
         #: Store the extension register info
         self.extension = extra_config.pop("extension", None)
         self.extension_reg = extra_config.pop("extension_reg", None)
```
EmilioPeJu, line 266: is `extension_reg` actually useful? or is it just a relic from old code? I see it is used in `fmc_acq427` but I don't see any other code needing it, so we might get rid of it

EmilioPeJu:
Sorry, I didn't fully understand the server part handling the extension, it actually implements the read extension part.
There is still a small mistake in which the read registers are not being read in the auto-generated code but I think you already spotted that

common/python/configs.py
EmilioPeJu: I had a deeper look and I think we can get read of all appearances of `extension_reg`... it is redundant because when we specify `extension` we are already telling that it's not a real hardware register... we can see that the `system` block doesn't use it and it works fine

tomtrafford: I'm going to have a look at this, I remember previously having a look and coming to the conclusion that extension_reg was necessary. But I can't remember why now. If it isn't needed I'll remove it, if it is needed I will add some documentation on what it is doing....

common/templates/config.jinja2
```diff
@@ -17,11 +17,13 @@
     {% for field in block.fields %}
     {# field is a FieldConfig object #}
     {# insert its name and type (including subtype and options) #}
+        {% if field.no_config != 1 %}
```
EmilioPeJu, line 20: now that `no_config` is `bool` you can drop the `!= 1`

common/templates/descriptions.jinja2
```diff
@@ -5,7 +5,9 @@
     {% for field in block.fields %}
     {# field is a FieldConfig object #}
     {# insert its name and description #}
+        {% if field.no_config != 1 %}
```
EmilioPeJu, line 8: ditto

common/templates/registers.jinja2
```diff
@@ -9,7 +9,9 @@
     {% for field in block.fields %}
     {# field is a FieldConfig object #}
     {# ask it for its register addresses within the block base address #}
+        {% if field.no_config != 1 %}
```
EmilioPeJu, line 12: ditto

common/python/configs.py
tomtrafford: I believe `extension_reg` is used when the field is also a hardware register. This is described in the [documentation](https://pandablocks.github.io/PandABlocks-FPGA/master/reference/block.html#field-sections) - although it looks like I've removed the line as part of this pull request. The `fmc_acq427` block needs hardware registers for some of its extension fields.  The `system` block doesn't need hardware registers so doesn't use `extension_reg`. I think this is still the case and `extension_reg` is still needed?

EmilioPeJu: Just wanted to reiterate what we talked about earlier:
- yes, `extension_reg` creates a harware register, but the server doesn't know about it and doesn't use it
- in the specific case of `fmc_acq427`, those created registers are used for right shifting the output (i.e. multiplying by a power of 2), however, because it is never set by the server, it uses the default value of 0 and shifting by 0 leaves the same value intact
- I don't understand why we would want to do that right shifting (considering the source data is 32 bits), but if we really need that feature, we could easily do that inside the extension

EmilioPeJu: In summary, the old `extension_reg` is deprecated by the type `write_extension` and the extension returns the value to write

PR #94, opened 2022-11-07
Fix PCAP hdl test error and modify PCAP Sum^2 timing sections

shu-soleil:
- Fix PCAP hdl test error
- Modify Sum^2 timing sections in PCAP module for doc plot and hdl test


shu-soleil:
To be merged with changes of PR #92

PR #96, opened 2022-12-08
add OUT_MODE param to counter for rate meter application

shu-soleil:
add OUT_MODE param enum to COUNTER 
- On_change to reflect OUT on internal counter value change
- On_disable to reflect OUT on ENABLE falling edge


modules/counter/counter.timing.ini
```diff
@@ -39,6 +39,26 @@ scope: counter.block.ini
...
+24      : TRIG=0
+26      : ENABLE=0          -> OUT=18
+29      : ENABLE=1
+30      : TRIG=1
+31      : TRIG=0
+36      : TRIG=1
+37      : TRIG=0
+43      : TRIG=1
+44      : TRIG=0
+46      : ENABLE=0          -> OUT=14
```
coretl, line 60: I don't understand why this isn't 18?

coretl:
@tomtrafford is currently looking at getting the other PRs merged so we can see if the HDL tests pass or fail in CI. Do the HDL tests for counter work for you locally?

shu-soleil:
> @tomtrafford is currently looking at getting the other PRs merged so we can see if the HDL tests pass or fail in CI. Do the HDL tests for counter work for you locally?

Yes all HDL tests passed for Counter locally

modules/counter/counter.block.ini
```diff
@@ -14,6 +14,12 @@ description: Rising edge ticks the counter up/down by STEP
 type: bit_mux
 description: Up/Down direction (0 = Up, 1 = Down)
 
+[OUT_MODE]
+type: param enum
+description: Counter OUT update mode (on internal counter value change or on ENABLE falling edge)
+0: On_change
+1: On_disable
```
Araneidae, line 21: Think I'd probably prefer enum strings with spaces to be more consistent with other enumeration options: `On Change` and `On Disable`.

Araneidae (approved):
Looks ok, I think.

Have to confess I find the documentation a bit perplexing, don't really think I understand what this option is doing, but then the counter is already pretty complex.

modules/counter/counter.block.ini
shu-soleil: good with On-Change + On-Disable like LUT

coretl: That sounds fine to me, we have a mix of standards anyway

shu-soleil:
Could this PR be merged ?

Issue #95, opened 2022-11-21
Add RateMeter mode for Counter module


shu-soleil:
To add RateMeter feature mode to the COUNTER module, I propose the changes shown on the "featureCounterRateMeter" branch which implicates:

- add a new param MODE and a new bit_mux RELOAD
- the default MODE will be "0: Normal" which performs continuous counting as usual
- for all the modes, internal counter will be reset to START value on each RELOAD rising edge
- new MODE "RateMeter": COUNTER.OUT is latched and updated to the counter value on each RELAOD rising edge
- new MODE "Scaler": COUNTER.OUT is latched from ENABLE rising edge et updated on ENABLE falling edge

Would this be good for you ?

coretl:
I think the functionality is reasonable, but I'd like to see the requirements for the entire application to see if we can rationalize the bit_mux inputs. We have `TRIG` and `DIR`, and we can probably repurpose `DIR` to act as `RELOAD` if we think about it.

A sample usecase that I can think of:
- To have a DAC value output proportional to the count rate of a TTL input.

Could you give the use cases for your two new modes?

shu-soleil:
Do you use DIR bit of COUNTER? How about to keep RELOAD as but_mux and change DIR to enum param? 
Was DIR designed for pulse/dir counting ? In this case we could add PulseDir mode to QDEC module for this purpose.

COUNTER's new modes are to replace our CPCI counting card.
The most usecase is pulse counting for detectors as: point detector, intensity monitor (+VFC), photomultiplier (+discriminator) etc., in step-scan or free running monitoring.

coretl:
We have the same use case for point detectors and intensity monitors. The way we do it is:
- For captured data, set `COUNTER.OUT.CAPTURE=Diff`, then set the counter with `MIN=0` and `MAX=0`. Each trigger signal will give the difference between the counter at the beginning and end of the gate. So if you had a 1Hz trigger, you would get a counts per second reading each second.
- For free running, we use `FILTER.MODE=difference`, then trigger it at 1Hz. Then `FILTER.OUT` will give a counts per second.

Does this solve your use cases?

shu-soleil:
- For captured data we do it in the same way as you
[...]

PR #97, opened 2022-12-08
add WIDTH param to CLOCK for issue #95

shu-soleil:


coretl:
The logic and tests look fine to me.
@Araneidae or @glennchid would you mind reviewing the VHDL please?

Araneidae (approved):
On the whole looks good and simple to me.

I would perhaps recommend one change, just to ease potential timing pressure.  Instead of computing `full_period` and `low_period` in variables I'd suggest assigning them to registers to simplify the subsequent counter and output logic.  Of course, you'd then also need to register the `reset` flag so that you're resetting with up to date values.

modules/clock/hdl/clock.vhd
```diff
@@ -33,40 +33,42 @@ end clock;
...
         end if;
         -- if WIDTH=0 then set OUT high time to half period
         if (unsigned(WIDTH) = 0) then
             high_period := '0' & full_period(31 downto 1);
         else
             high_period := unsigned(WIDTH);
         end if;
-        low_period := full_period - high_period;
-        
+        low_period <= full_period - high_period;
```
Araneidae, line 68: I'm a bit concerned about this change: `full_period` and `high_period` are now changed on difference clock cycles, which means there will be one clock when there's a discrepancy.

Maybe this handled properly by the timing of the reset, I see that reset updates in the same tick as `full_period`, and is used to initialise `counter32`, and `low_period` won't be used during this doubtful tick.

I think the figure below shows the potential issue:
timing

shu-soleil: `counter32` is updated one clock behind `reset` or `enable` or `pull_period`, so it's aligned with `low_period` for good comparison I think.

shu-soleil:
Could this PR be merged ?

modules/clock/hdl/clock.vhd
tomtrafford: @Araneidae before I merge this PR I just wanted to check that you were happy with this change, or are you still concerned?

Araneidae: @tomtrafford , yes, I was happy with Shu's explanation.  It's one of these issues where it's easy to make a subtle mistake, and the timing figure shows that it is on the _edge_ of being wrong ... but is ok!

PR #98, opened 2023-01-13
Change IP build to global synthesis rather than out-of-context (OOC)

glennchid:
I was experimenting with global synthesis for IP (synthesising the IPs together with the main FPGA logic) as opposed to OOC which we currently use. OOC is the default in Vivado and the recommended flow by Xilinx, as it avoids need to  rebuild the IPs for any changes on the logic, but global synthesis provides the possibility of better optimisation. Although in this case I couldn't tell whether it gave better quality of results or not, I noticed a significant reduction in build time for a single zpkg, contrary to expectations.

The times below are on my desktop for the PandABox-no-fmc app:

| | OOC | Global synth | Comment |
| --- | --- | --- | --- |
| `time make carrier_ip` | 17:14 | 1:50 | Time to create IP files |
|`time make ps_core` | 1:32 | 1:34 | Time to create PS core file (already using global synth) |
|`time make` |  22:12 | 22:41 | Time to build just FPGA zpkg |
|`make clean-all`| - | - | Delete all output products |
|`time make`| 40:59 | 25:46 | Time to build IP, ps, and zpkg from scratch |
|`make clean`| - | - | Delete FPGA zpkg (leaving IP and PS files) |
|`time make`| 22:13 | 21:48 | Recheck time to recreate just FPGA zkpg |
 
To me it seems worthwhile making this change. Just wanted to get other opinions on this.


valerixb:
Hi Glenn. Interesting find: thank you for sharing! It's kind of surprising; maybe we are dominated by the time spent setting up the big database at the beginning and saving the results at the end of IP compilation; even if we are more than 80% full, the overall compile time for Panda is ~1 hour, while Xilinx efforts are probably developed with ~24h compilation tasks in mind; what do you think? Whatever the reason, it's a great improvement in compiling time for us, so I totally think it's worth doing it! Thank you!

valerixb (approved):

glennchid:
@valerixb I suspect you are right that for PandA our IPs are so small that it takes around the same amount of time to read in the DCPs in OOC mode as out does to synthesise them in the first place. Presumably this would be different for more complex IPs.

coretl (approved):

shu-soleil (approved):

shu-soleil:
Very interesting change. I've tested a 'make' after 'clean-all' with your changes, it takes now 23 minutes very close to that with OOC. And global synthesis seems really improving the optimisation, it loosens a bit the timing path constraints which is very useful. Good job

EmilioPeJu (approved):

PR #99, opened 2023-02-03
Remove app files from targets to avoid duplication

EmilioPeJu:


Araneidae:
Given that this PR simply removes numerous files without any other comment, it would be good to at least know where the originals of these duplicated files are.  I think the commit message could usefully include this information, which would help further review.

Araneidae (approved):
Looks sensible to me

Issue #82, opened 2022-07-01
Move Jinja defs out of top_defines


coretl:
Make a new small top_defines_gen.vhd.jinja2 which contains just 

```
-- Bit Bus Width, Multiplexer Select Width -------------------------------
constant BBUSW                  : natural := 128;
constant BBUSBW                 : natural := 7;

-- Position Bus Width, Multiplexer Select Width.
constant PBUSW                  : natural := {{ total_pos_bus_length }};
constant PBUSBW                 : natural := 5;

-- Extended Position Bus Width.
constant EBUSW                  : natural := 12;
--------------------------------------------------------------------------

-- Presence of PCAP_STD_DEV functionality
constant PCAP_SUPPORTS_STD_DEV  : std_logic := '{{ ("pcap_std_dev" in fpga_options) | int}}';
```

Make another package and load it from top_defines

adhowell:
I'm not sure this is possible with the VHDL package system. E.g. if `top_defines` uses `top_defines_gen`, then the contents of `top_defines_gen` are only accessible in the scope of `top_defines`. VHDL does not have a proper linker system that would make the contents of `top_defines_gen` accessible to a file that only uses `top_defines`.

coretl:
@Araneidae I believe you suggested this, do you know if it is possible?

Araneidae:
If there are only a small number of symbols that need to be defined in `top_defines_gen` the simplest trick is to re-export them from `top_defines` thus:
```vhdl
use work.top_defines_gen.all;
package top_defines is
    -- These definitions have to be autogenerated
    alias PBUSW is work.top_defines_gen.PBUSW;
    alias PCAP_SUPPORTS_STD_DEV is work.top_defines_gen.PCAP_SUPPORTS_STD_DEV;

    -- All the rest go here
    ...
end;
```

PR #101, opened 2023-02-10
Added top_defines_gen template and made top_defines non-templated

adhowell:


common/hdl/top_defines.vhd
```diff
@@ -1,7 +1,11 @@
+-- AUTOGENERATED
```
tomtrafford, line 1: Probably should remove the autogenerated header, as this file no longer is. I think this file should be moved to the defines directory as well?

common/python/generate_app.py
```diff
@@ -301,8 +301,8 @@ def generate_soft_blocks(self):
                              "soft_blocks.vhd")
         self.expand_template("addr_defines.vhd.jinja2", context, hdl_dir,
                              "addr_defines.vhd")
-        self.expand_template("top_defines.vhd.jinja2", context, hdl_dir,
-                             "top_defines.vhd")
+        self.expand_template("top_defines_gen.vhd.jinja2", context, hdl_dir,
```
tomtrafford, line 304: Can you also make sure this file gets generated for the testbenches. Maybe at the same time as the module.tcl? You may need to generate some dummy parameters for total_pos_bus_length

adhowell:
Seeing as the makefile always generates the app I have just added the generated top_defines_gen.vhd path to both the tcl scripts. I'm not 100% sure this is a good idea though as this won't work in an instance where the test scripts are ever completely separated from the app generation...

tomtrafford:
This will resolve #82

Issue #88, opened 2022-10-05
Add PCAP interrupt on PCAP.ENABLE high when armed


coretl:
As per https://github.com/PandABlocks/PandABlocks-server/issues/7, when ARMed, we would like PCAP to raise an interrupt to the server when ENABLE rises so the server can record an NTP timestamp that corresponds to captured time=0

PR #110, opened 2023-03-02
Add PCAP interrupt on PCAP.ENABLE high when armed and enabled

EmilioPeJu:
This change provide a new interrupt source when PCAP becomes active and enabled, if this condition happens in isolation, the block will not be switched (which was the default in every interrupt before this change).

Additionally, a driver compatibility register was added which is checked when the driver is loaded.

This fixes https://github.com/PandABlocks/PandABlocks-FPGA/issues/88


EmilioPeJu:
This change comes together with https://github.com/PandABlocks/PandABlocks-server/pull/34

PR #113, opened 2023-03-13
Clocking changes to allow fine delay and recovered clock

glennchid:
Having convinced myself that it is enough to declare the clocks as asynchronous groups with the `-include_generated_clocks` flag (given that the clocks are async and we don't make use of any advanced CDC constraints such as `max_delay -datapathonly`), I eventually conceded and re-added exclusive constraints for the multiplexed clocks (although not-strictly necessary) to appease the `report_cdc` tool. I decided to just use the auto-derived names of these clocks, rather than renaming them with `create_generated_clock` as this causes complications with the recovered clock name and the processing order of the constraint files by Vivado.

Also, for clean living, added synchronisers to some MGT derived signals which are used in the clock mux selection, and renamed clock signals for better clarity.

### DIFF
--- a/modules/system/system.block.ini
+++ b/modules/system/system.block.ini
@@ -76,7 +76,7 @@ scale: 0.001494582
 type: read
 description: PLL locked for SMA external clock
 
-[EXT_CLOCK]
+[CLOCK_SOURCE]
 type: param enum
 description: External sma and event receiver clock enables
 0: int clock
--- a/modules/system/system.block.ini
+++ b/modules/system/system.block.ini
@@ -81,7 +81,7 @@ type: param enum
 description: External sma and event receiver clock enables
 0: int clock
 1: sma clock
-2: event receiver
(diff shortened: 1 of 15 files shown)


modules/system/system.block.ini
```diff
@@ -76,12 +76,12 @@ scale: 0.001494582
...
 description: PLL locked for SMA external clock
 
-[EXT_CLOCK]
+[CLOCK_SOURCE]
 type: param enum
 description: External sma and event receiver clock enables
 0: int clock
 1: sma clock
-2: event receiver
+2: mgt recovered clock
```
tomtrafford, line 84: Change from mgt recovered clock to sfp1 recovered clock

glennchid:
Converted to draft pending resolution of issue #119

modules/system/system.block.ini
glennchid: Done!

PR #114, opened 2023-03-14
Fix for USB on PandABox

glennchid:
Fix for USB on PandABox for PandABlocks v3 (Vivado 2020 and later).

With some help from @EmilioPeJu and the Xilinx wiki:
https://xilinx-wiki.atlassian.net/wiki/spaces/A/pages/18842272/Zynq+Linux+USB+Device+Driver
we added a node for the usb-phy to the devicetree.

The corresponding change for the xu5 was not found to be necessary, as the USB already works fine with v3. Not sure about the ZedBoard but from recollection this did not work in the past, though USB support is not so useful for this target.

### DIFF
--- a/targets/PandABox/target-top.dts
+++ b/targets/PandABox/target-top.dts
@@ -14,6 +14,13 @@
 		device_type = "memory";
 		reg = <0x0 0x40000000>;
 	};
+	usb_phy0: phy0 {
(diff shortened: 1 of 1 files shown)


EmilioPeJu (approved):
Looks good

PR #116, opened 2023-03-27
Changes to build system to only generate IP required for App

glennchid:
* Allows for adding optional IP repos external to Vivado, so not all sites need to generate all IP for a particular target
* Does timestamp/dependency checking of each IP rather than build-once for each target as before, but will only regenerate the new IP when building new app
* At the expense of slightly longer runtime for IP product generation (c. 3-4 mins for no_fmc app, as opposed to 1-2 mins for all PandABox IPs) due to overhead from launching vivado multiple times (if anyone has an idea how to do this better, please suggest)
* Another side-effect is that `make edit_ips` will no longer show all IPs for a target, just those required for the app selected in CONFIG, but `make carrier_ip` can be run again for the other apps to fully populate the IP project.

### DIFF
--- a/CONFIG.example
+++ b/CONFIG.example
@@ -12,6 +12,9 @@ VIVADO_VER = 2020.2
 
 # Definitions needed for FPGA build
 export VIVADO = /scratch/Xilinx/Vivado/\$(VIVADO_VER)/settings64.sh
+# Path to external IP repository (if required)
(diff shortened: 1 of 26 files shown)


EmilioPeJu (approved):
Looks good to me

EmilioPeJu (approved):

valerixb (approved):

PR #117, opened 2023-03-31
Vivado2022.2 upgrade

glennchid:
- Decided to remove BOARD files for the hardware targets as these are not needed in our build flow, and are increasingly more difficult to maintain with newer versions of vivado. For instance, the ZedBoard BSP files are no longer distributed as part of Vivado but need to be checked out from the xhub board_store, or the alternative would be to keep the vendor's files in our repo - neither seems a good solution.
- U-boot build now requires RHEL8 or above, due to libcrypto and rsa-sign dependencies.

Note, CI will fail for this branch until the tool versions are updated.


valerixb (approved):

EmilioPeJu (approved):
Looks good to me, I don't think the board part were used at all.
The board files contains: the fpga part (which we have always defined), some constrains (which we have always provided under `targets/<target>/const`) and some PS block default configuration (which we  have always configured inside  `targets/<target>/bd/panda_ps.tcl`)... therefore, it's good to remove those board files.

glennchid:
Converted to draft pending resolution to issue #119

glennchid:
When running in a container Vivado now requires `glibc-langpack-en` under Rocky8 or `locales` package under ubuntu  (and maybe some extra configuration as well), to set the locale.

PR #120, opened 2023-04-11
Autogen fixes

EmilioPeJu:
- Move version and constant files to autogen target
- Use python specified in CONFIG for building constants file

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,7 +163,7 @@ $(VERSION_FILE) : $(VER)
 
 
 $(CONSTANT_FILE) : $(TOP)/common/templates/registers_server
-	$(TOP)/common/python/generate_constants.py "$<" > $@
+	$(PYTHON) $(TOP)/common/python/generate_constants.py "$<" > $@
 
 # ------------------------------------------------------------------------------
 # Documentation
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -234,7 +263,7 @@ FPGA_TARGETS = fpga-all fpga-bit carrier_fpga carrier_ip ps_core \
                fsbl devicetree boot u-boot dts xsct sw_clean u-boot-src \
                dtc atf ip_clean ps_clean
 
-$(FPGA_TARGETS): $(TOP)/common/fpga.make $(AUTOGEN_BUILD_DIR) | update_VER
```
glennchid, line 237: Do we not still update_VER to be an order-only prerequisite?

Makefile
```diff
@@ -244,7 +273,7 @@ else
 	$(MAKE) -C $(FPGA_BUILD_DIR) -f $< VIVADO_VER=$(VIVADO_VER) \
         TOP=$(TOP) TARGET_DIR=$(TARGET_DIR) APP_BUILD_DIR=$(APP_BUILD_DIR) \
         TGT_BUILD_DIR=$(TGT_BUILD_DIR) TOP_MODE=$(TOP_MODE) DEP_MODE=$(DEP_MODE) \
-		VER=$(VER) $@
```
glennchid, line 247: $(VER) is also passed to FPGA.make as part of `run_sim_%` target, line 207

PR #127, opened 2023-05-10
Remove number in *METADATA.LABEL for a block with only a single entry

### LINKED ISSUE #126: There should be no number in *METADATA.LABEL for a block with only a single entry
E.g. `*METADATA.LABEL_PCAP` not `*METADATA.LABEL_PCAP1`

tomtrafford:
Closes #126

### DIFF
--- a/common/templates/config.jinja2
+++ b/common/templates/config.jinja2
@@ -5,9 +5,13 @@
     EXPORTS             multiline
 {% for block in server_blocks %}
     {# Make a label for holding the block description #}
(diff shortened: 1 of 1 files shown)


coretl:
Actually, I just realized that this will break Malcolm, and we don't really have a nice introspection mechanism for backwards compatibility. Can we leave it as it is?

tomtrafford:
Sure, I will close this. But, it's an easy change to implement if we want to bring it back in the future.

PR #130, opened 2023-06-01
Add double table sequencer feature

EmilioPeJu:
A table ending with a zeroed entry indicates that there will be a next table, a new table can be pushed as soon as CAN_WRITE_NEXT becomes 1.
The user is responsible to push a table that gives enough time to be able to push the next table before the last line is reached.
The last table will not contain the last zeroed entry and will be the only table that will be repeated according to the REPEATS register.

I added tests that cover the main use case and the most important failure cases.


EmilioPeJu:
I found that there are timing problems in some cases, I will think about how to improve that. The other failed test is about the python timing simulation, `seq_sim.py` needs to be updated... I'm making this a draft.

EmilioPeJu:
The critical timing paths are in the pcap, the routine becomes difficult because of the high utilization, I was able to "fix" the failing apps by disable pcap_dev_std in those, I don't think there is much that can be done about it (other than moving to the next generation panda)

Araneidae:
It'd be good to have a closer look at some of the failing paths.  PCAP in particular is a good candidate for pipelining, and there may be some useful clues in the specifics of the path failing timing.

EmilioPeJu:
Looking at the failing build (before my last commit), we found that the failed path is actually in the CALC block, doing 4 sums and a shift in the same clock ticks might be too much, Michael suggests that doing that in 2 clock ticks will help (depending on how important is the 1 tick requirement that it initially had)

coretl:
I would prefer to keep the 1 tick timing but reduce the functionality of the CALC block. @Araneidae suggested reducing SHIFT to a maximum value of 2 and writing as a case statement to make it more efficient. Would this work?

EmilioPeJu:
Yes, that worked, I added that change and removed the commit disabling pcap_std_dev option and timing constraints seems to be met, the build time is still very long but that is expected with such high utilisation

modules/calc/calc.block.ini
```diff
@@ -50,7 +50,7 @@ description: Scale divisor after add
 [SHIFT]
```
Araneidae, line 50: Don't forget to remove the `[FUNC]` block as well; @coretl tells me it's a relic of an earlier version.

modules/calc/hdl/calc.vhd
```diff
@@ -63,8 +63,11 @@ begin
     inpd_in <= convert(inpd_i, TYPED(0));
     calculation <= inpa_in + inpb_in + inpc_in + inpd_in;
 
-    -- Shift result as required
-    output <= shift_right(calculation, to_integer(unsigned(SHIFT(4 downto 0))));
+    -- Shift result as required, this is limited to range 0-2 to reduce
+    -- FPGA resources
+    output <= calculation when SHIFT = x"00000000" else
+              "0" & calculation(32 downto 0) when SHIFT = x"00000001" else
+              "00" & calculation(31 downto 0);
```
Araneidae, line 70: Huh.  That does work.  Might be nicer to put this calculation in the process, if more long winded:
```vhdl
case SHIFT(1 downto 0) is
    when "00" => out_o <= calculation(31 downto 0);
    when "01" => out_o <= calculation(32 downto 1);
    when "10" => out_o <= calculation(33 downto 2);
    when others => out_o <= (others => '-');
end case;
```
Not sure if the don't care would help with synthesis...

EmilioPeJu: trying that now, sorry about my dyslexic stupid mistake doing the shifting

modules/calc/calc.block.ini
EmilioPeJu: done

modules/calc/hdl/calc.vhd
EmilioPeJu: strangely, the no-fmc app didn't pass timing, while all the other apps (which adds more modules) passed, this indicates we are in a situation in which the non-deterministic optimization in vivado might pass or fail randomly

coretl: no-fmc has 2x SFP sync, so is still pretty full...

common/hdl/defines/top_defines.vhd
```diff
@@ -138,6 +138,14 @@ record
     time2       : unsigned(31 downto 0);
 end record;
 
+constant SEQ_ZERO : seq_t := (repeats => (others => '0'),
+                              trigger => (others => '0'),
+                              out1 => (others => '0'),
+                              out2 => (others => '0'),
+                              position => (others => '0'),
+                              time1 => (others => '0'),
+                              time2 => (others => '0'));
```
Araneidae, line 147: You can probably write this as `(others => (others => '0'))`!

Also, where is this constant used?  Does it earn its keep as a global constant?  Does it maybe belong in `sequencere_defines.vhd` instead?

modules/calc/hdl/calc.vhd
```diff
@@ -63,13 +61,16 @@ begin
...
     process (clk_i) begin
         if rising_edge(clk_i) then
-            out_o <= std_logic_vector(output(31 downto 0));
+            -- Shift result as required, this is limited to range 0-2 to reduce
+            -- FPGA resources
+            case SHIFT(1 downto 0) is
+                when "00" => out_o <= std_logic_vector(calculation(31 downto 0));
+                when "01" => out_o <= std_logic_vector(calculation(32 downto 1));
+                when "10" => out_o <= std_logic_vector(calculation(33 downto 2));
+                when others => out_o <= (others => '-');
```
Araneidae, line 72: Might be neater to make `calculation` a `std_logic_vector` and do the conversion when computing the sum.  Saves all the conversions in the case statement anyway.

common/hdl/defines/top_defines.vhd
EmilioPeJu: I tried initializing that way and didn't work (not sure if due to vivado or vhdl), but anyway, I don't think the constant use justified its existence, so I've removed it, thanks for spotting that

EmilioPeJu:
Judging by the last builds, I think when we pass the timing, we are marginally passing it and any random change could make it fail again, should we maybe disable `pcap_std_dev` in the applications that are more likely to fail timing?

tests/sim/sequencer_double_table/bench/sequencer_double_table_tb_compile.tcl
```diff
@@ -0,0 +1,17 @@
...
+create_project $TEST $BUILD_DIR/build/tests/$TEST -force -part xc7z030sbg485-1
+
+set_property top $TEST [current_fileset -simset]
+add_files $TOP/common/hdl/
+add_files [glob $TOP/modules/seq/hdl/*.vhd]
+add_files $TOP/tests/sim/$TARGET/bench
+set_property FILE_TYPE "VHDL 2008" [get_files *.vhd]
+
+launch_simulation
+restart
```
glennchid, line 16: `restart' should not be necessary - this was the function causing the sims to crash with Vivado 2022.2, though strangely it has been working okay for me recently with some other code.

modules/calc/hdl/calc.vhd
```diff
@@ -61,15 +59,18 @@ begin
...
     process (clk_i) begin
         if rising_edge(clk_i) then
-            out_o <= std_logic_vector(output(31 downto 0));
+            -- Shift result as required, this is limited to range 0-2 to reduce
+            -- FPGA resources
+            case SHIFT(1 downto 0) is
+                when "00" => out_o <= calculation(31 downto 0);
+                when "01" => out_o <= calculation(32 downto 1);
+                when "10" => out_o <= calculation(33 downto 2);
+                when others => out_o <= (others => '-');
```
glennchid, line 72: do we really 'not care' what the output is when SHIFT is set to a value higher than 2 (even though it is limited in the server the FPGA does not know this)? Not sure how this will synthesise.
Would `when others => null;' be clearer?

modules/seq/hdl/sequencer_defines.vhd
```diff
@@ -0,0 +1,35 @@
+library ieee;
+use ieee.std_logic_1164.all;
+use ieee.numeric_std.all;
+
+library work;
```
glennchid, line 5: library `work' doesn't need to be declared, but it doesn't do any harm either.

modules/calc/hdl/calc.vhd
Araneidae: We really don't care, and I'm hoping that this gives the synthesiser an opportunity to simplify (though I'm inclined to bet that it's just going to generate zeros in this case).  Writing `null` would add a clock enable ... though that would be pretty cheap, and should provide the same optimisation opportunity.

coretl, line 72: How about treating a shift>=2 as a shift of 2?

```suggestion
                when others => out_o <= calculation(33 downto 2);
```

Araneidae: Of course, that's also perfectly sensible.  Now, do we want to paint the bikeshed blue or green?

coretl: Blue of course, it's the best colour by far...

modules/seq/hdl/sequencer_defines.vhd
EmilioPeJu: Removed!

tests/sim/sequencer_double_table/bench/sequencer_double_table_tb_compile.tcl
EmilioPeJu: Done

modules/calc/hdl/calc.vhd
EmilioPeJu: Done as you suggested in the code, the block ini already imposes a high limit, so this might as well be considered "undefined behavior", not sure on how much improvement in optimization the  'don't care' part was providing, I hope not noticeable.

EmilioPeJu:
Just let you know that I tested it in the hardware and it worked as expected, including the error cases of overruning or underruning with the new tables.

PR #132, opened 2023-06-23
Counter module: add TRIG_EDGE param

shu-soleil:
add TRIG_EDGE param for Counter module to select counting trigger edge (same as the Pulse module)


modules/counter/hdl/counter.vhd
```diff
@@ -119,36 +135,39 @@ begin
...
-            -- Count up/down on trigger
-            -- Initialise next_counter with current value
-            next_counter := resize(signed(counter),next_counter'length);
-            -- Direction
-            if (dir_i = '0') then
-                next_counter := next_counter + signed(STEP_default);
-            else
-                next_counter := next_counter - signed(STEP_default);
+        elsif (enable_i = '1' and trigger_edge = '1') then
+            if (counter_carry = '1') then
```
EmilioPeJu, line 139: We don't need this condition block, you can just set counter_carry to 0 unconditionally

modules/counter/hdl/counter.vhd
```diff
@@ -119,36 +135,39 @@ begin
...
-                counter_carry <= '1';
-                carry_latch <= '1';
-                next_counter := next_counter + signed(MAX_VAL - MIN_VAL + 1);
+            if (got_trigger = '1') then
+                -- Count up/down on trigger
+                -- Initialise next_counter with current value
+                next_counter := resize(signed(counter),next_counter'length);
+                -- Direction
+                if (dir_i = '0') then
+                    next_counter := next_counter + signed(STEP_default);
```
EmilioPeJu, line 149: Regarding the `signed` conversions, I suggests we make counter and STEP_default be `signed` and that way we can drop all the `signed` conversions scattered everywhere

EmilioPeJu: The same applies to `MAX_VAL`, `MIN_VAL` ... etc

modules/counter/hdl/counter.vhd
shu-soleil: But counter_carry need to be '1' if overflow occurs, otherwise set to 0

modules/counter/hdl/counter.vhd
shu-soleil: I didn't make changes on this part from the original version, but yes we can do so

shu-soleil:
Hi @EmilioPeJu, could you have a look at the CI checks for PandABox related Zpkg builds ? At least 2 of them passed on my local machine but all of them have failed here.

modules/counter/hdl/counter.vhd
EmilioPeJu: Sorry I didn't explain, you are setting it to 0 if it is 1, and `counter_carry` can only be 0 or 1, which means the result is always having it to 0, that's why you can just set it to 0 unconditionally

EmilioPeJu: when the overflow condition occurs, the assignment setting it to 1 will override the 0

shu-soleil: Emm not sure if my understand of your question is right, this means counter_carry will be '0' if it was '1' at previous clock tick.

EmilioPeJu:
Hi @shu-soleil , I have re-run your test, the problem was that the container git command was failing to obtain the version, I've released a new version of the container with that fixed and should now work

modules/counter/hdl/counter.vhd
EmilioPeJu: Sorry, that was not a question, what I meant is that you can replace:
```vhdl
if (counter_carry = '1') then
-- Need to stop the counter_carry on next trigger edge
counter_carry <= '0';
end if;
```

with just:
```vhdl
-- Need to stop the counter_carry on next trigger edge
counter_carry <= '0';
```

and the result in this case will be the same

coretl: So no matter if `counter_carry` is 0 or 1, at the end of either of the code blocks Emilio posted, it will be set to 0. They are both equivalent, so might as well use the second one which is simpler.

shu-soleil: Sorry I get it now, I'll update it

EmilioPeJu (approved):
Looks good to me, thanks for this change and for improving the existing code

shu-soleil:
Hello all tests have passed, could this be merged to the master ? Thanks

PR #138, opened 2023-07-25
Sync protocol changes

glennchid:
Major change to PandASync protocol

* 8 bit-bus entries are transmitted each clock tick
* Packing as follows:

| BITS[7:0] | POS1[31:16] | K28.5  |
| BITS[7:0] | POS1[15:0]  | POS2[31:24] |
| BITS[7:0] | POS2[23:0]  |
| BITS[7:0] | POS3[31:8]  |
| BITS[7:0] | POS3[7:0]   | POS4[31:16] |
| BITS[7:0] | POS4[15:0]  | CHECKBYTE   |

CHECKBYTE currently holds either all zeros, or a sequence number (which is checked for consecutive values). It could also hold some kind of checksum but does not currently.

Remaining issues:
* Link still requires resetting manually after initialisation
* Bogus(?) protocol-error detected after plugging fibre, requires manual reset to clear

I tried adding a settling-time delay to hold the MGT in reset until after the link has stabilised. This has no effect for small delays (<10 microsec), but for delays >100 microsec the MGT never comes out of reset. This approach works on other designs with similar MGTs.

### DIFF
--- a/modules/sfp_panda_sync/hdl/sfp_panda_sync_mgt_interface.vhd
+++ b/modules/sfp_panda_sync/hdl/sfp_panda_sync_mgt_interface.vhd
@@ -12,7 +12,7 @@ entity sfp_panda_sync_mgt_interface is
 
     port (GTREFCLK          : in  std_logic;
           SYNC_RESET_i      : in  std_logic;
-          sysclk_i             : in  std_logic;
+          sysclk_i          : in  std_logic;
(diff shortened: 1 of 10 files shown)


tomtrafford:
@Araneidae would you be able to review this PR please?

EmilioPeJu (approved):

PR #139, opened 2023-08-01
Simplify axi_lite_slave and make it faster

EmilioPeJu:
The following states were removed:
- State READ_START from the reading state machine
- State WRITE_START from the writing state machine

This achieves a simpler state machine and it is 1 clock faster, because now, the strobe signal is asserted in the transition from idle

### DIFF
--- a/common/hdl/axi_lite_slave.vhd
+++ b/common/hdl/axi_lite_slave.vhd
@@ -16,12 +16,12 @@
 -- selected by "module select" is pulsed for one clock cycle after the
 -- write_address and write_data outputs are valid:
 --
---  State           | IDLE  | START |WRITING| DONE  |
---                           ________________________
---  write_data_o,   XXXXXXXXX________________________
+--  State           | IDLE  |WRITING| DONE  |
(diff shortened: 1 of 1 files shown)


EmilioPeJu:
This has been tested in a real PandA, though it would still be nice having a test bench

Araneidae:
This is sound, but I do have one big reservation: this is a protocol change, and there may be some old code buried in there somewhere that makes use of the early presentation of the address.

To summarise: in the original code, the read or write address was being presented one tick before the corresponding strobe.  I do remember that before the ack was plumbed into the design this extra tick was being used in some places to implement register reads.

I'm optimistic that there cannot be too many places where this kind of behaviour can be hiding, as most of the PandA registers are implemented in a very uniform way.  I do think however that a more careful check of the more obscure register destinations is needed before this is merged.  Specifically, look out for register reads where there is no delay on the ack.

EmilioPeJu:
We checked the special cases for register's access (mainly PCAP) and they are fine

Issue #158, opened 2023-09-20
EVR: Vary MGT refclk frquency


glennchid:
On-board oscillators can be controlled via I2C from the slow FPGA.
A feedback circuit could be implemented to track the system clock frequency, whether this is coming from the PS fabric or an MGT recovered clock.

PR #160, opened 2023-09-20
Sync picxo

### LINKED ISSUE #104: Add Picxo to Sync

glennchid:
Adding PICXO IP for synchronous data transfer between PandAs. Works by synchronising MGT transmit clocks to the system clock. See https://docs.xilinx.com/v/u/en-US/xapp589-VCXO for more details.

Added to both PandA sync and EVR apps. Tested and working for PandA sync. Implemented for EVR, however due to offset in frequency between the event clock (derived from the machine RF) and the MGT refclk oscillators, the event clock is too far out of range for the PICXO to lock. See #158.

Requires PICXO IP sources to be available to Vivado.

Closes #104


Makefile
```diff
@@ -319,9 +318,7 @@ ZPKG_DEPENDS += $(APP_BUILD_DIR)/extensions
 ZPKG_DEPENDS += $(DOCS_HTML_DIR)
 
 $(APP_BUILD_DIR)/ipmi.ini: $(APP_FILE)
-	$(PYTHON) -m common.python.copy_file_in_modules \
-        --fallback $(TOP)/common/templates/default_ipmi.ini \
-        $(TOP) $< ipmi.ini $@
+	$(PYTHON) -m common.python.make_ipmi_ini $(TOP) $< $@
```
EmilioPeJu, line 321: I renamed `make_ipmi_ini` in master to make it reusable and to pull out the content of default `ipmi.ini` to a file (making it easier to maintain), not sure why this is changed in this pull request

common/hdl/defines/picxo_pkg.vhd
```diff
@@ -0,0 +1,128 @@
...
+        CE_PI2_O : OUT STD_LOGIC;
+        CE_DSP_O : OUT STD_LOGIC;
+        OVF_PD : OUT STD_LOGIC;
+        OVF_AB : OUT STD_LOGIC;
+        OVF_VOLT : OUT STD_LOGIC;
+        OVF_INT : OUT STD_LOGIC
+      );
+    END COMPONENT;
+
+    COMPONENT picxo_ila
```
EmilioPeJu, line 56: Do we want to keep debugging stuff in here?

modules/sfp_eventr/hdl/sfp_receiver.vhd
```diff
@@ -93,6 +115,22 @@ attribute ASYNC_REG of EVENT4_WSTB_meta2 : signal is "TRUE";
 attribute ASYNC_REG of event_bits_meta1  : signal is "TRUE";
 attribute ASYNC_REG of event_bits_meta2  : signal is "TRUE";
 
+type event_code_array is array(natural range <>) of std_logic_vector(7 downto 0);
+constant event_codes: event_code_array(0 to 27) := (x"20", x"21", x"23", x"24", x"25", x"26",
```
EmilioPeJu, line 119: isn't this limiting the usefulness of the event receiver? why do we consider only those codes as good ones? additionally, the value that we get when we receive a COMMA character is also in that list, is that a coincidence?

ip_defs/sfp_panda_sync.tcl
```diff
@@ -52,3 +52,11 @@ CONFIG.gt0_val_clk_cor_seq_2_3 {00000000}               \
 CONFIG.gt0_val_clk_cor_seq_2_4 {00000000}               \
 ] [get_ips sfp_panda_sync]
 
+proc patch {BUILD_DIR} {
+    puts "Patching generated IP sources"
```
EmilioPeJu, line 56: am I right in thinking that this is needed to make the gtwizard IP compatible with the PICXO IP?

CONFIG.example
```diff
@@ -13,7 +13,7 @@ VIVADO_VER = 2022.2
 # Definitions needed for FPGA build
 export VIVADO = /scratch/Xilinx/Vivado/\$(VIVADO_VER)/settings64.sh
 # Path to external IP repository (if required)
-#export EXT_IP_REPO =
+export EXT_IP_REPO = /scratch/Xilinx/XAPP589/PICXO_FRACXO_v32_2
```
EmilioPeJu, line 16: Could we have this IP available in a shared location under `/dls_sw/...`

coretl: We need to make sure this is installed on the S3 bucket too...

modules/sfp_eventr/hdl/sfp_receiver.vhd
EmilioPeJu: OK, I see my confusion was about the name, it's more a `valid_first_byte` than an event code

CONFIG.example
glennchid: Yes, this is the path within the container. I will add to /dls_sw too..

modules/sfp_eventr/hdl/sfp_receiver.vhd
```diff
@@ -27,18 +27,36 @@ entity sfp_receiver is
...
+    probe_in0 : IN STD_LOGIC_VECTOR(31 DOWNTO 0);
+    probe_in1 : IN STD_LOGIC_VECTOR(31 DOWNTO 0);
+    probe_in2 : IN STD_LOGIC_VECTOR(31 DOWNTO 0);
+    probe_in3 : IN STD_LOGIC_VECTOR(31 DOWNTO 0);
+    probe_in4 : IN STD_LOGIC_VECTOR(31 DOWNTO 0);
+    probe_in5 : IN STD_LOGIC_VECTOR(31 DOWNTO 0);
+    probe_out0 : OUT STD_LOGIC_VECTOR(0 DOWNTO 0)
+  );
+END COMPONENT;
+
```
Araneidae, line 51: Should we be including debug artefacts in merged code?

Makefile
glennchid: oops, yes I see now how this got dropped. Yes, it caused an error - but hopefully fixed now!

common/hdl/defines/picxo_pkg.vhd
glennchid: It won't consume resource as its just the component definition included in the package. It can be deleted though - I just wanted to make it easier to put the chipscope interfaces back in if something stops working.

ip_defs/sfp_panda_sync.tcl
glennchid: Yes, that's right. We need to modify the sources that are generated by the IP. It may be possible to put this information directly into the IP tcl script, which would be neater, but I don't know.

modules/sfp_eventr/hdl/sfp_receiver.vhd
glennchid: Probably not, unfortunately the EVR is still a bit of a work-in-progress, see comment in PR and subsequent tickets.

PR #162, opened 2023-10-06
Remove over-riding of ctrl block ack ports where feasible

glennchid:
* write_ack now set in ctrl block and not wrapper
* read_ack over-riding removed for hardware blocks
* read_ack added to modules which did not have one previously

The only blocks which treat the `acks` differenently are `system` due to the slow FPGA registers, and `pcap` due to the DRV block.

NB: I checked this built for all zpkgs and tested one zpkg on hardware and it worked fine, but just noticed this is failing the python tests on PR. I think I can fix this before merging.


glennchid:
Moved DRV acks for PCAP inside the pcap_core_ctl (non-autogenerated) for consistency, so pcap is now less of a special case.

EmilioPeJu (approved):
Looks good to me

Araneidae (approved):
Looks good to me as well.  This looks like a pretty thorough clean-up of the acknowledge handling, definitely needs to be merged!

PR #163, opened 2023-10-11
Update FPGA boot files release

tomtrafford:
These changes update the release process for the FPGA boot files created by a 'make boot' command.

- The created devicetree.dtb and boot.bin are now contained within a zip file tagged with the target name and the git version.
- The zip file is now uploaded as an artifact as part of the CI and is included as an asset on release.
- There is a git version included in the devicetree.dtb. The rootfs should be able to read this and display the version to the user.


targets/PandABox/target-top.dts
```diff
@@ -80,3 +80,9 @@
 		};
 	};
 };
+
+/ {
+	pandainfo {
+		version = "GIT_VERSION";
+	};
+};
```
coretl, line 88: @EmilioPeJu where would you put this in the device tree?

EmilioPeJu: An alternative approach would be hanging version from the root node, for example:
```
/ {
	model = "PandABox";
	version = "GIT_VERSION";
	....
}```
The approach of adding a pandainfo node is also acceptable, specially if we plan to add more information properties in future.
In practices, the only thing that changes is if you read `/proc/device-tree/version` or `/proc/device-tree/pandainfo/version` to get the value.

PR #174, opened 2024-02-07
Vivado2023 Upgrade

glennchid:
Vivado 2023.2 upgrade

Issues encountered:

- Vivado build consistently seg faulting when run inside container. Related to licensing bug: https://support.xilinx.com/s/question/0D54U00005Sgst2SAB/failed-batch-mode-execution-in-linux-docker-running-under-windows-host?language=en_US. Workarounds: run vivado with `LD_PRELOAD` pointing to `libudev.so.1` or pass through host networking stack into container (`--net-host`) - the latter is preferable if it works under CI.
- IRQ sensitivity defaulting to disallowed value for zynqmp - all targets changed to EDGE_RISING for consistency
- Kernel build failed when using previous config for zynqmp - removing unused features fixes build
- Verilog testbenches need renaming to .sv for changes to verilog 2-d array handling in vivado
- Bug introduced into fscanf function in verilog, failing to parse multiple negative numbers: https://support.xilinx.com/s/question/0D54U000080bmlESAQ/sscanf-and-fscanf-broken-in-vivado-20232-when-parsing-negative-numbers?language=en_US
- For unknown reason, labels for temperature and voltage sensors in sysfs changed for zynqmp device, but not for zynq-7000. I think the new labels correspond to the the same sensors that were previously monitored. There seem to be several voltage and temperature sensors available, so we may want to review the optimal choice for zynqmp.
- Devicetree change for SD card that was found to be necessary for `PandABrick` with earlier version of vivado (https://github.com/PandABlocks/PandABlocks-FPGA/commit/6f27fba69f762de02547a98059d2b8e1c715bbfb) was not found to be necessary for `xu5_st1`. May have been a consequence of faulty SD card or cage on ST1, or different version of vivado/kernel.

NB: Draft PR subject to all sites being ready to upgrade. ~~Merge needs to be made in concert with PR in PandA-rootfs repo:
https://github.com/PandABlocks/PandABlocks-rootfs/pull/44~~


CONFIG.example
```diff
@@ -1,23 +1,23 @@
...
-# commented out can be left as they are, the default value is show.  Uncommented
+# commented out can be left as they are, the default value is shown.  Uncommented
 # lines must be assigned values.
 
 # Default build location.  Default is to build in build subdirectory.
 BUILD_DIR = /build
 
 # Development Tool Version
-VIVADO_VER = 2022.2
+# VIVADO_VER = 2023.2
```
EmilioPeJu, line 11: Why is this now commented?

Makefile
```diff
@@ -18,7 +17,7 @@ MAKE_ZPKG = $(PANDA_ROOTFS)/make-zpkg
 MAKE_GITHUB_RELEASE = $(PANDA_ROOTFS)/make-github-release.py
 
 BUILD_DIR = $(TOP)/build
-VIVADO_VER = 2022.2
+VIVADO_VER = 2023.2
```
EmilioPeJu, line 20: I see the example CONFIG suggests to set this variable, if this is a default one, shouldn't it be set with `?=` after including CONFIG the same way we do with `VIVADO_MODE`?

CONFIG.example
glennchid: Just to make it consistent with the comments at the top of the file. According to this it should only be the uncommented variables that need to be set in the CONFIG, the others have default values in the Makefile.

Makefile
glennchid: Not sure it matters for these, as CONFIG is read after and overwrites them. Probably either way is the same.

EmilioPeJu (approved):

valerixb (dismissed):

tomtrafford (approved):

Issue #172, opened 2024-02-05
Add support for hardware derived timestamp


coretl:
At the moment the user is presented with 2 things in [data capture](https://pandablocks.github.io/PandABlocks-server/master/capture.html)
- `arm_time`: NTP accurate timestamp when server received the `*PCAP.ARM` message
- `start_time`: NTP accurate timestamp when kernel driver received interrupt from FPGA on rising edge of `PCAP.ENABLE`

We would like to add support for timestamps for `start_time` to be derived from the EVR transmitted time.

The suggested interface at FPGA level:
- `SFP_output_interface` gains two extra 32-bit entries, `TS_SEC` and `TS_NSEC`
- EVR block populates these with the timestamp, all other SFP blocks leave them at zero
- `SYSTEM` has a new `TIMESTAMP_SOURCE` register that allows user to select between None, SFP1, SFP2 or SFP3, and wires the correct SFP TS registers to a single top level `TS_SEC` and `TS_NSEC`
- Add new `*REG.PCAP_TS_SEC` and `*REG.PCAP_TS_NSEC` registers, and get `PCAP` to latch from top level `TS_SEC` and `TS_NSEC` into them on rising edge of `ENABLE`

The suggested interface at server level:
- Change `hw_get_start_ts` to generate not just `ts`, but `hw_ts` which comes from `*REG.PCAP_TS_SEC` and `*REG.PCAP_TS_NSEC`
- Pass this to `send_capture_info` as `pcap_hw_start_tsp`
- If it is non-zero then use it to produce `start_time` in the header, and also add a `hw_time_offset` which is `pcap_start_tsp - pcap_hw_start_tsp`

The suggested interface at IOC level:
- Continue writing timestamps as relative
- Add new HDF attributes at top level for `arm_time`, `start_time`, and `hw_time_offset`

EmilioPeJu:
Done in PandABlocks/PandABlocks-FPGA#177 and PandABlocks/PandABlocks-server#45

PR #177, opened 2024-02-20
Adding FPGA support for hardware timestamp from SFPs

glennchid:
See issue #172 and PR in https://github.com/PandABlocks/PandABlocks-server/pull/45
N.B. feature only implemented for the PandABox target, as SFP support still to be added for the Ultra-scale targets.


EmilioPeJu (approved):
Looks good to me

EmilioPeJu (approved):

Araneidae (approved):

PR #181, opened 2024-03-07
Don't waste time in `pcap_dma`

EmilioPeJu:
There is no reason to not start transfering to memory if we have more than 1 burst worth of data in the fifo.


Araneidae (approved):
Looks good to me

Issue #167, opened 2023-11-08
Investigate mechanism to read built-in MAC addresses from Secure EEPROM on XU5 and XU6


glennchid:
Under petalinux u-boot reads MAC address from the Secure EEPROM and populates this in the device-tree <local_mac_address> field. In our build a random locally administered addresses is put in the device-tree by u-boot, which is later over-ridden in linux with the value in `/qspi/MAC` via `ifconfig`. The present work-around is to manually write the Enclustra-assigned MAC address to the QSPI.

I tried the following entry in the dts (copied from the petalinux generated dts):
```
&gem0 {
	local-mac-address = [ff ff ff ff ff ff];
};
```
but this did not work on its own, as more config is required in petallinux, see eg:
https://support.xilinx.com/s/question/0D52E00006hpQdiSAE/uboot-ethernet-mac-address-from-eeprom-petalinux-20202?language=en_US

EmilioPeJu:
There are 2 pull requests dealing with this: https://github.com/PandABlocks/PandABlocks-FPGA/pull/186 and https://github.com/PandABlocks/PandABlocks-rootfs/pull/49

EmilioPeJu:
Solved in the PRs I mentioned in my last comment

PR #186, opened 2024-03-15
Allow each target to define the way to obtain its MAC address

EmilioPeJu:
At the moment only 2 targets define it:
PandABox: MAC from a qspi memory
PandABrick: MAC from a secure EEPROM

This is part of a solution for issue PandABlocks/PandABlocks-FPGA#167 and it's related to https://github.com/PandABlocks/PandABlocks-rootfs/pull/49

### DIFF
--- a/common/fpga.make
+++ b/common/fpga.make
@@ -99,7 +99,7 @@ carrier_ip: $(APP_IP_DEPS)
 ps_core: $(PS_CORE)
 devicetree : $(DEVTREE_DTB)
 fsbl : $(FSBL)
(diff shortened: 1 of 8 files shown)


EmilioPeJu:
I know this opens up more refactoring work by pulling more target specific configuration out of rootfs, however, I would like to stick to the original issue and deal with the rest in a separate ticket

EmilioPeJu:
I also added a change to use `target-defs` for getting the target specific information that led-daemon needs

Araneidae (approved):
Looks sensible to me

Araneidae (approved):

PR #187, opened 2024-03-28
Github Actions CI Changes

shihab-dls:
CI workflow jobs now spread out into reusable workflows. Some repetitive steps (i.e., checkouts, rclone installations, S3 mounting, and registry login) moved to a composite actions. S3 mounting and registry login removed from actions, however, due to secrets becoming visible as inputs.

PR opened as draft until IRIS runners are available for CI jobs.


Araneidae:
Please squash these commits into a sensible commit history!  I suspect a lot of the churn in this will be due to debugging history, which doesn't need to go into the final PR.

shihab-dls:
All jobs are now on IRIS runners, but make_boot and test jobs require more attention:
- make_boot currently hangs between successful build completion and artefact upload (for 4hrs). A 60min timeout and zip file check is set to appropriately end the job.
- Running tests on IRIS take ~3.5x longer than on GH runners, with each test (of 161) sometimes requiring a couple of minutes.

The /scripts directory has been deleted, and an NFS mount at /opt/arc-panda on cs05r-sc-cloud-30 (rw for DCS group) is used instead of the S3 bucket. An adjusted dev-container is used as the runner's base image, a PR of which will be opened on the rootfs repository.

shihab-dls:
A work-around for the 4hr test job is now in place
- The /scripts directory was reintroduced
- Two scripts are included: Find_Tests.sh and Group_Tests.py
--Find_Tests.sh produced an array of module names and their respective test counts (i.e., number of tests)
--Group_Tests.py allocates modules to IRIS runners, targeting runners with least load first, resulting in a (almost) even spread of tests.

Overall duration of the matrix of test jobs is < 1hr; thus, no longer the limiting job. Overall duration of complete workflow is ~1hr.

tomtrafford (approved):

coretl (approved):
A skim of this looks reasonable

.github/workflows/_make_boot.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      # Generate bootable image
+      - name: Make boot
+        id: make_boot
+        run: |
+          echo "boot_files=false" >> $GITHUB_OUTPUT
+          cd repos/PandABlocks-fpga
+          ln -s CONFIG.example CONFIG
+          make boot APP_NAME=${{ matrix.app }}
+          ps -Ao pid= -o comm= | awk '$2 ~ /Xvfb/ { print $1}' | xargs kill
```
coretl, line 41: Comment would be good here

.github/workflows/_test_matrix.yml
```diff
@@ -0,0 +1,32 @@
...
+            path: repos/PandABlocks-fpga
+            fetch-depth: 0
+  
+        - name: Evaluate number of modules/tests & generate job matrix
+          id: make_matrix
+          run: |
+               bash $WORK_DIR/repos/PandABlocks-fpga/.github/scripts/Find_Tests.sh "$WORK_DIR/repos/PandABlocks-fpga/modules" "$WORK_DIR/repos/PandABlocks-fpga/.github/scripts"
+               json_content=$(cat github_tests.json)
+               echo "::set-output name=matrix::$json_content"
+               echo "$json_content"
```
coretl, line 32: Is this finding of tests and separating into different runners via the matrix needed with the cocotb tests?

.github/workflows/_undo.yml
```diff
@@ -0,0 +1,19 @@
+on:
+  workflow_call:
+
+jobs:
+  undo:
+    runs-on: ubuntu-latest
+    if: ${{ always() && failure() && github.event_name == 'push' && contains(github.event.head_commit.message,'undo_true')}}
+    steps: 
+      - name: Checkout
```
coretl, line 9: When does this run? Needs a comment to say

coretl (approved):
Couple of comments, just needs comments adding I think then fine to merge

.github/workflows/_test_matrix.yml
tomtrafford: No, this was done to reduce the time taken for the testbenches. It was taking ~4 hours for all of the tests to complete as Vivado's simulation was single threaded, so @shihab-dls made a script to separate them into 4 roughly equivalent sized batches of tests so they can be run in parallel on different runners. However, the whole cocotb test only takes a few minutes.

.github/workflows/_undo.yml
tomtrafford: This reverts a merge, if after merging it fails CI. I think there was some discussion a few months ago whether this should be automatic or not and it was decided to add a condition that the text "undo_true" is in the merge request title. 

However we actually use the merge request titles for Changelog entries now, so may not be the best place to add this... Is there a better place to put this flag. If this job isn't just automatic, I'm not certain it would be used... @coretl do you have thoughts on this?

.github/workflows/_test_matrix.yml
coretl: Ok, happy to merge this as is then simplify later with the cocotb changes

.github/workflows/_undo.yml
coretl: Personally I would delete this functionality, then "undo" the merge using the GitHub UI if it failed after merge...

coretl (approved):

PR #189, opened 2024-04-11
Update CONFIG.example

tomtrafford:
Remove \


tomtrafford:
If I print VIVADO in the Makefile I can see it is set to "/dls_sw/FPGA/Xilinx/Vivado/\2023.2/settings64.sh" but doesn't seem to affect functionality.

EmilioPeJu (approved):

PR #191, opened 2024-05-01
xu5: show network status on the OLED display

EmilioPeJu:
This change only affect target xu5_st1 but can be easily ported to others. The display controller is a SSD1306.

fbpad(a frame buffer terminal emulator) is expected to by provided by the rootfs.

### DIFF
--- a/targets/xu5_st1/target-top.dts
+++ b/targets/xu5_st1/target-top.dts
@@ -3,10 +3,19 @@
 
 &sdhci1 {
 	disable-wp;
+	no-1-8-v;
 };
 
 &i2c0 {
 	clock-frequency = <100000>;
+	ssd1306: oled@0x3c {
+		compatible = "solomon,ssd1306";
+		reg = <0x3c>;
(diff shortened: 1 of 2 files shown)


targets/xu5_st1/target-top.dts
```diff
@@ -3,10 +3,19 @@
 
 &sdhci1 {
 	disable-wp;
+	no-1-8-v;
```
glennchid, line 6: the 'no-1-8-v' is a strange one, as I had previously noticed that it didn't seem to need to be included when using Vivido-2023.2, despite being needed before: https://github.com/PandABlocks/PandABlocks-FPGA/pull/174

glennchid (approved):
Looks fine to me :)

Araneidae (approved):

targets/xu5_st1/target-top.dts
EmilioPeJu: That's strange, I built master in both repos and I needed this fix

PR #192, opened 2024-06-06
Allow capability of adding additional SFP ports through FMC card.

tomtrafford:
Changes to use the FMC MGT as additional connections.

Restructures FMC and SFP VHDL interfaces using View (VHDL 2019).

The Loopback app is setup to have a 4th SFP sync using these changes. This can then be used by using an FMC card with additional SFP ports.

TODO: Add documentation showing how to edit target.ini and block.ini files to use this functionality.


glennchid (approved):

targets/xu5_st1/xu5_st1.target.ini
```diff
@@ -2,7 +2,9 @@
 # module. The config files for each carrier block can be found in
 # blocks/blockname
 [.]
-io: fmc: 1, i, o, io
+io: fmc: 1
+    fmc_mgt: mgt, 0
+additionalMGT: 4
```
tomtrafford, line 7: fmc_mgt: mgt, 4*

EmilioPeJu (approved):
Looks good, the interface feature improves code readability.
One thing I noticed is that for the FMC MGTs the MAC address passed is 00:00:00:00:00:00, which I suppose is not a problem for the use-case we want, if at some point we need to set that, we might want to extend the register interface to allow passing more than 4 MAC addresses.

Issue #179, opened 2024-02-29
PCAP.TS_START not zeroed when GATE already high at the start of capture


coretl:
If `PCAP.GATE` and `PCAP.ENABLE` are both tied to `ONE`, like in this diagram:
![image](https://github.com/PandABlocks/PandABlocks-FPGA/assets/101418278/a3a0c597-02dd-4412-9f8a-4172998e0f37)

then `TS_START` on the last frame will be leftover from the previous scan rather than being 0:

```
$ nc 172.23.252.201 8889

OK
arm_time: 2024-02-29T17:31:03.969Z
start_time: 2024-02-29T17:31:03.970Z
missed: 0
process: Scaled
format: ASCII
fields:
 PCAP.TS_START double Value scale: 8e-09 offset: 0 units: s
 PCAP.TS_END double Value scale: 8e-09 offset: 0 units: s

 0 0.500000064
 0.500000064 1.500000064
 1.500000064 2.500000064
 2.500000064 3.500000064
 3.500000064 4.500000064
END 5 Disarmed
arm_time: 2024-02-29T17:31:10.377Z
start_time: 2024-02-29T17:31:10.377Z
missed: 0
process: Scaled
format: ASCII
fields:
 PCAP.TS_START double Value scale: 8e-09 offset: 0 units: s
 PCAP.TS_END double Value scale: 8e-09 offset: 0 units: s

 4.889428296 0.500000064
 0.500000064 1.500000064
 1.500000064 2.500000064
 2.500000064 3.500000064
 3.500000064 4.500000064
END 5 Disarmed
```

tomtrafford:
I've added a timing test in [pcap-TS_START-test ](https://github.com/PandABlocks/PandABlocks-FPGA/tree/pcap-TS_START-test). However, I cannot replicate this behaviour in the VHDL. @EmilioPeJu Does a change need to be made in the server similar to https://github.com/PandABlocks/PandABlocks-server/issues/37 ?

EmilioPeJu:
The relevant change to the FPGA already happened when the start time feature was added. Was this bug introduced after that? if you don't see the problem in the timing tests, it might be a hardware problem that needs to be investigated with chipscope (or similar)

EmilioPeJu:
It would be nice simplifying that code, though that would require more work(and testing) than just fixing this bug

tomtrafford:
I've tried putting these changes onto a PandA but it does not seem to have fixed this problem.

[...]

PR #196, opened 2024-06-28
WIP: capture ts_start on rising enable if gate is active

EmilioPeJu:
This is a proposal to fix https://github.com/PandABlocks/PandABlocks-FPGA/issues/179, based on Tom's investigation


EmilioPeJu:
Closing this PR because it didn't fix the problem

PR #197, opened 2024-07-18
counter: add register to set counter value

EmilioPeJu:
- Setting the counter take precedence over a trigger event, that way the block will output the set value at least for one tick
- It is possible to set a value while the block is disabled (this allows to control the pos bus value without having to actually count)
- Timing tests were added for two cases: setting a value while enabled and setting a value while disabled
- Python counter simulation script was updated to consider the set register

### DIFF
--- a/modules/counter/counter.block.ini
+++ b/modules/counter/counter.block.ini
@@ -28,6 +28,11 @@ description: Counter OUT update mode (on internal counter value change or on ENA
 0: On-Change
 1: On-Disable
 
+[SET]
+type: param int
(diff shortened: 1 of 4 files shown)


glennchid (commented):
What is the use case for the SET value? Should it have priority over the enable signal, i.e. if SET value or provided, then ENABLE, should the counter count from ST or from START?

glennchid (approved):
Looks fine, provided we are happy with the behaviour of SET vs START.

Issue #204, opened 2024-08-29
Connect Step/Direction signals on PandABrick encoders development branch


tomtrafford:
Currently the, the STEP and DIRECTION signals of the pmacenc block have been disabled. They need to be enable, to do this:

- Un-comment their fields in pmacenc.block.ini (modules>>pmacenc). The autogen framework will automatically add them to the ctrl blocks.
- Connect the signals from the pmacenc_ctrl block instantiation in PandABrick_encoders_block.vhd (targets>>PandABrick>>hdl) to the PandABrick_encoders instantiation. And add them to the PandABrick_encoders port map (PandABrick_encoders.vhd).
- The signals are on the bit bus so they need to be added in [PandABrick_top.vhd ](https://github.com/PandABlocks/PandABlocks-FPGA/blob/020f13d022fdbeaab3f4789296ac333088c62a86/targets/PandABrick/hdl/PandABrick_top.vhd#L1303C1-L1304C75)(targets>>PandABrick>>hdl). This will need wiring through PandABrick_encoders_top.vhd.
- These will need to replace the existing signals: step, dir within the PandABrick_encoders.vhd

PR #216, opened 2024-10-10
Pandabrick encoders

### LINKED ISSUE #184: PandABrick encoders
For the PandABrick the encoders will be split into three blocks: INCENC, ABSENC and PMACENC as shown in the below drawing.
PandABrick Encoders

tomtrafford:
Changes to split PandABrick encoders into ABSENC, INCENC and PMACENC. Step Direction on PMACENC still needs to be implemented and an issue has been made for this #204.

Resolves #184

### DIFF
--- a/targets/PandABrick/hdl/PandABrick_encoders.vhd
+++ b/targets/PandABrick/hdl/PandABrick_encoders.vhd
@@ -67,7 +67,7 @@ port (
     SETP_i              : in  std_logic_vector(31 downto 0);
     SETP_WSTB_i         : in  std_logic;
     RST_ON_Z_i          : in  std_logic_vector(31 downto 0);
-    STATUS_o            : out std_logic_vector(31 downto 0);
+    STATUS_o            : out std_logic;
     INCENC_HEALTH_o     : out std_logic_vector(31 downto 0);
     HOMED_o             : out std_logic_vector(31 downto 0);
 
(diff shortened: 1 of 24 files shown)


targets/PandABrick/hdl/PandABrick_encoders.vhd
```diff
@@ -23,10 +23,10 @@ port (
     posn_i              : in  std_logic_vector(31 downto 0);
     enable_i            : in  std_logic;
     -- Encoder I/O Pads
-    INENC_A_o           : out std_logic;
-    INENC_B_o           : out std_logic;
-    INENC_Z_o           : out std_logic;
-    INENC_DATA_o        : out std_logic;
+    -- INCENC_A_o          : out std_logic;
+    -- INCENC_B_o          : out std_logic;
+    -- INCENC_Z_o          : out std_logic;
```
glennchid, line 28: Are we no longer providing ABX signals onto the bitbus?

targets/PandABrick/hdl/PandABrick_encoders.vhd
glennchid, line 1: This file is very long and hard to navigate. Is it possible to split it up and have separate hierarchy for the INCENC, ABSENC and PMACENC? Or at least break the file down into different logical sections?

glennchid (approved):

targets/PandABrick/hdl/PandABrick_encoders.vhd
tomtrafford: With there being 8 encoders, there became too many signals on the bit_bus. Tom told me that there is not currently a use-case for using these signals so they could be removed.

targets/PandABrick/hdl/PandABrick_encoders.vhd
tomtrafford: There are three sections designated by comments such as `-----------------------------INCENC---------------------------------------------` for the different blocks. I'm happy to try to break it down further if you think that would be helpful, but the PMACENC overlaps functionality with both the INCENC and ABSENC which causes some issues with separating them.

Araneidae: I would suggest breaking the processing in this file into a number of sub-entities, maybe called `encoders_incenc`, `encoders_absenc`, `encoders_pmacenc` (I've not tried to make sense of it).  The act of creating such a breakdown, although often challenging, can be very instructive in itself: finding the correct boundaries between entities is the key to a good design.  There are a lot of signals here, so finding the correct breakdown might be tricky.

glennchid (approved):

PR #218, opened 2024-10-21
Updated documentation for adding additional io

### LINKED ISSUE #193: Write Documentation for using SFP through FMC interface
Write documentation for explaining the how to use the new functionality in #192

tomtrafford:
resolves #193.

@glennchid Would you say this description is easy to follow? I'm wondering if it doesn't read that well - I'm happy to re-write it.

### DIFF
--- a/docs/reference/block.rst
+++ b/docs/reference/block.rst
@@ -60,6 +60,8 @@ this:
     ip:
     otherconst:
     extension:
+    extra_interface: fmc_mgt: mgt 4
+
 
 The ``description`` should be a short (a few words) description that will be
 visible as a Block label to users of the `pandablocks_device_` when it runs.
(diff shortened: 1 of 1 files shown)


docs/reference/block.rst
```diff
@@ -284,12 +292,20 @@ The first entry to the ini file defines information for the SFP sites for the
 target::
```
glennchid, line 292: Also, change 'SFP' to 'IO' in 
```
The [.] section
~~~~~~~~~~~~~~~
The first entry to the ini file defines information for the SFP sites for the
```

glennchid (dismissed):
A few minor suggestions..

glennchid:
> resolves #193.
> 
> @glennchid Would you say this description is easy to follow? I'm wondering if it doesn't read that well - I'm happy to re-write it.

I think it is okay, though a few minor suggestions above. Might be better though to get somebody not familiar with the changes to review and see if it is understandable.

PR #229, opened 2025-05-09
Implement streaming table feature for modules seq and pgen

EmilioPeJu:
This change allows seq and pgen to get the table data via DMA even at the same time the table is being used.
In order to efficiently implement driver notification, a new interrupt was introduced and a register that is cleared when read, this register(`TABLE_IRQ_STATUS`) is a bitset with 2 bits per instance, set on 2 events: when the table becomes ready to accept a new buffer and when all the transfers were completed.
The reason why `TABLE_IRQ_STATUS` is part of `REG` block is that `DRV` is mostly pcap stuff and taking that out from pcap would require a significant plumbing work which I don't think is justified.
This should be merged together with server change PandABlocks/PandABlocks-server#63.

For more information, see issue PandABlocks/PandABlocks-server#61.

Diagram and report https://github.com/EmilioPeJu/panda-server-issue-61-streaming-table

### DIFF
--- a/.github/workflows/_make_boot.yml
+++ b/.github/workflows/_make_boot.yml
@@ -48,3 +48,4 @@ jobs:
         with:
           name: boot-${{ matrix.app }}
           path: /build/boot*.zip
+          if-no-files-found: error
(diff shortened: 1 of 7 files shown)


Araneidae (dismissed):
I think this is ready to be merged

PR #230, opened 2025-05-12
Add a vhdl PID for the PandA

DiamondRC:
Created a PID block for use on PandA devices.

The PID accepts four parameters from input signals, allowing for custom tuning to fit the system. These include: a proportioal term, a derivative term, an integral term and a feed-forward term.

The PID takes a setpoint value and a measured value and returns a correction term for motion.

Currently the PID caps values at the signed limits, which are dynamically calculated from the size of the input/output signals. Whilst this provides flexability, if the values are fixed then this could be simplified to conserve board space.

### DIFF
--- a/modules/pid/hdl/pid.vhd
+++ b/modules/pid/hdl/pid.vhd
@@ -1,0 +1,108 @@
+--------------------------------------------------------------------------------
+--  File:       PID.vhd
+--  Desc:       HDL implementation of a PID
+--------------------------------------------------------------------------------
+
+library ieee;
+use ieee.std_logic_1164.all;
+use ieee.std_logic_signed.all;
+use ieee.numeric_std.all;
+
+entity PID is 
+port (
+    clk_i             : in  std_logic; -- Forced by PandA
+    init_i            : in  std_logic;
+    slo_clk_i         : in  std_logic;
+
+    -- Pass in constants externally
+    Kp_i              : in  std_logic_vector(31 downto 0);
+    Ki_i              : in  std_logic_vector(31 downto 0);
+    Kd_i              : in  std_logic_vector(31 downto 0);
+    FF_i              : in  std_logic_vector(31 downto 0);
+     
+    setpoint_i        : in  std_logic_vector(31 downto 0); -- Desired value
+    in_signal_i       : in  std_logic_vector(31 downto 0); -- Measured value
+    out_signal_o      : out std_logic_vector(31 downto 0)  -- Output value
+);
+end entity PID;
+
+architecture Basic_PID of PID is
+    signal in_error   : signed(31 downto 0) := (others => '0');
+    signal prev_error : signed(31 downto 0) := (others => '0'); -- Start from zero
+    signal integral   : signed(63 downto 0) := (others => '0'); -- Start from zero
+    signal derivative : signed(31 downto 0) := (others => '0');
+    signal P_out      : signed(63 downto 0) := (others => '0');
+    signal I_out      : signed(63 downto 0) := (others => '0');
+    signal D_out      : signed(63 downto 0) := (others => '0');
+    signal round_out  : signed(63 downto 0) := (others => '0');
+
+    -- For type conversion
+    signal Kp         : signed(31 downto 0) := (others => '0');
+    signal Ki         : signed(31 downto 0) := (others => '0');
+    signal Kd         : signed(31 downto 0) := (others => '0');
+    signal FF         : signed(31 downto 0) := (others => '0'); -- Feed Forward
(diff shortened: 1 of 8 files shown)


modules/pid/hdl/pid.vhd
```diff
@@ -0,0 +1,108 @@
...
+                FF         <= (others => '0');
+            else
+                -- Proportional
+                P_out      <= (Kp * in_error);
+
+                -- Integral
+                I_out      <= I_out + (Ki * in_error);
+
+                -- Derivative
+                D_out      <= D_out + (Kd * (in_error - prev_error));
```
LeandroMartinsdS, line 81: Shouldn't be

`D_out <= (Kd * (in_error - prev_error));`
?

As it is, from my point-of-view, it will add a kind second proportional term

modules/pid/hdl/pid.vhd
```diff
@@ -0,0 +1,108 @@
...
+
+                -- Protect against overflow
+                if P_out + I_out + D_out + FF > max_lim then
+                    report "OVERFLOW HIGH";
+                    round_out  <= (round_out'left => '0', others => '1');
+                elsif P_out + I_out + D_out + FF < min_lim then
+                    round_out  <= (round_out'left => '1', others => '0');
+                    report "OVERFLOW LOW";
+                else
+                    round_out  <= P_out + I_out + D_out + FF;
```
LeandroMartinsdS, line 94: I would modify to
`round_out <= P_out + I_out + D_out + FF_out;`

being 

`FF_out <= (FF * setpoint);`

modules/pid/hdl/pid.vhd
```diff
@@ -0,0 +1,108 @@
...
+                end if;
+
+            end if; -- Reset + Logic
+        end if; -- Clock
+    end process;
+
+    -- Error calc
+    in_error     <= setpoint - in_signal;
```
LeandroMartinsdS, line 102: Shouldn't the `in_error` be updated inside the process?

modules/pid/hdl/pid.vhd
```diff
@@ -0,0 +1,108 @@
...
+    setpoint      <= signed(setpoint_i);
+    in_signal     <= signed(in_signal_i);
+
+    process(slo_clk_i)
+        variable max_lim : signed(out_signal'range);
+        variable min_lim : signed(out_signal'range);
+
+    begin
+        max_lim := (out_signal'left => '0', others => '1');
+        min_lim := (out_signal'left => '1', others => '0');
```
LeandroMartinsdS, line 64: does `max_lim` and `min_lim` need to be redefined at every process iteration?
Couldn't they be constants outside the process?

modules/pid/hdl/pid.vhd
DiamondRC: I think this would mean the derivative and integral term would need to wait until the next clock edge to use the in_error value? Keeping it outside the process allows the value to be immediately available.

modules/pid/hdl/pid.vhd
```diff
@@ -0,0 +1,109 @@
...
+    FF            <= signed(FF_i);
+    setpoint      <= signed(setpoint_i);
+    in_signal     <= signed(in_signal_i);
+
+    process(slo_clk_i)
+        constant max_lim : signed(out_signal'range) := (out_signal'left => '0', others => '1');
+        constant min_lim : signed(out_signal'range) := (out_signal'left => '1', others => '0');
+
+    begin
+        if rising_edge(slo_clk_i) then
```
Araneidae, line 64: `slo_clk_i` is an ordinary signal derived from the bit bus (this is clear from its definition below in `pid.block.ini`).  I'm afraid you simply cannot use this a "fabric" clock, which is what you're trying to do here.  Instead you need to write:
```vhdl
process (clk_i) begin
    if rising_edge(clk_i) then
        if slo_clk_i then
            ... do stuff ...
        end if;
    end if;
end process;
```
Things then get more tricky, you'll need to figure out just how quickly everything proceeds, and you might need to internally create 1 and 2 tick delayed versions of `slo_clk_i` to help manage your calculations.

Araneidae (changes requested):
A couple of extra thoughts:

- This needs to be marked as Draft as it has clearly never been synthesised
- There is no testbench, there really does need to be something

LeandroMartinsdS:
It might be worth to have a look at the https://github.com/PandABlocks/PandABlocks-FPGA/pull/188

EmilioPeJu:
CI is failing because master is now using ipk packages instead of zpg, you can fix it by rebasing your branch on top of master

PR #241, opened 2025-07-11
Refactored code from encoders.vhdl into inenc and outenc .vhdl

adedamola-sode:
Code from encoder.vhd has been split into inenc and outenc .vdl respectively. Tested on p99 panda, but needs validation.

### DIFF
--- a/common/hdl/encoders/outenc.vhd
+++ b/common/hdl/encoders/outenc.vhd
@@ -1,0 +1,245 @@
+library ieee;
+use ieee.std_logic_1164.all;
+use ieee.numeric_std.all;
+
+library work;
+use work.top_defines.all;
+use work.support.all;
+
+entity outenc is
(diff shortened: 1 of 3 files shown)


common/hdl/encoders/outenc.vhd
```diff
@@ -0,0 +1,245 @@
...
+IOBUF_Zs0 : entity work.iobuf_registered port map (
+    clock   => clk_i,
+    I       => Zs0_opad,
+    O       => Zs0_ipad,
+    T       => outenc_ctrl(0),
+    IO      => Zs0_pad_io
+);
+
+
+-- Determine how to share between inenc and outenc
```
tomtrafford, line 230: There's a couple of comments you've added in the file here... Does something still need to be done, or could you remove the comments.

PR #242, opened 2025-07-23
Panda2 ps wip

adedamola-sode:
Related to issue #236 

The basic ps has been created,but needs testing. This ticket is a work in progress.


EmilioPeJu:
Hi @adedamola-sode, I think the file `targets/PandABox2/bd/panda_ps.bd.tcl` is a mistake, it should be called `targets/PandABox2/bd/panda_ps.tcl` (which is also there but probably identical?)

EmilioPeJu (approved):
These changes are self-contained and provide support for softblocks only, it was tested on a ST1 and we expect to extend it gradually once there is a prototype.

PR #245, opened 2025-09-01
Include 2 PGEN instance

EmilioPeJu:
Apparently, some of the last refactoring changes have improved utilisation enough to include PGEN again.

### DIFF
--- a/includes/common_soft_blocks.include.ini
+++ b/includes/common_soft_blocks.include.ini
@@ -28,5 +28,8 @@ number: 4
 [SEQ]
 number: 2
 
+[PGEN]
+number: 2
+
 [SRGATE]
 number: 4
--- a/includes/pandabrick_soft_blocks.include.ini
+++ b/includes/pandabrick_soft_blocks.include.ini
@@ -28,5 +28,8 @@ number: 4
 [SEQ]
 number: 2
 
+[PGEN]
+number: 2
+
 [SRGATE]
 number: 4


EmilioPeJu:
@coretl , I don't think there is a use-case yet that requires 2 instances, Have you encountered some user with that requirement?
Regarding the instance in pandabox, I was basically thinking about I15, they need it for their new modulation feedback system, otherwise we would need to create and build one new application that includes one, Would you prefer that approach?

EmilioPeJu:
@coretl, I found that it fit, I've amended the commit

EmilioPeJu:
> > @coretl , I don't think there is a use-case yet that requires 2 instances, Have you encountered some user with that requirement? Regarding the instance in pandabox, I was basically thinking about I15, they need it for their new modulation feedback system, otherwise we would need to create and build one new application that includes one, Would you prefer that approach?
> 
> The problem is that when we go from 1 to 2 (or vice versa), we change the PVs that are created and whether the save files are valid or not. It also makes the ophyd interface difficult.

Why don't we fix the underlying problem? if I understood correctly, the problem is that from 1 to 2 instances, it changes the name from PGEN to PGEN1,in my opinion, it would be better to change the server to also use the name PGEN1 when there is only one instance, otherwise we have this excessive constraint of always needing more than 1 instance for every block with the consequently waste of resources

Araneidae:
One downside of my suggestion above is that `*BLOCKS?` no longer gives you a way to infer the true block names, and I can't think of a compatible fix for this.

EmilioPeJu:
> > @coretl, I found that it fits, I've amended the commit
> 
> Does it fit for PandABrick too?

I have just tested it,and it works too!

> I think at the moment this only affects the names in `*CHANGES?`, and for inputs both `PCAP` and `PCAP1` are treated as synonyms?

During testing for the streaming table, I remember setting the fields doesn't work on singletons addressed with the index, so I had to create a function to enumerate all blocks (based on *CHANGES? at start time) and take the first from the list that matches the block prefix.
If BLOCK and BLOCK1 could be treated as aliases in all situations(no matter how many instances there are), this could be a simpler fix that wouldn't break backward compatibility.

tomtrafford (approved):

PR #248, opened 2025-09-12
Remove zpkg building and dependency on rootfs

EmilioPeJu:
This commit was dropped from #244 to facilitate testing.
This change shouldn't be included in releases older than 5.0.


tomtrafford:
Also rename applications and target to be compatible with Yocto. And put a check to fail quickly if names are wrong.

PR #251, opened 2025-11-14
pcap: replace IP-based fifo in pcap_dma

EmilioPeJu:
This has 2 main advantages:
- Code is more portable
- We can now use any simulator

Tested in hardware without problems.


Araneidae:
Only one question: is any other part of the project still using IP `fifo_1K32_ft`?  If not, shouldn't this also be removed from the `ip_defs` directory?

EmilioPeJu:
@Araneidae , Happy with this?

PR #252, opened 2025-11-14
spbram: don't use shared variable for memory (not VHDL compliant)

EmilioPeJu:
A simple signal can be used instead, because this is used only in pcap_buffer, which is written when the capture is prepared and is read during the capture, therefore, the case in which the memory is read and written in the same tick doesn't matter.


EmilioPeJu:
In the last commit, I reduced the data width of the memory in pcap_buffer, given that the mask is only 10 bits, this might reduce the used resources in some targets.

Araneidae (approved):
That was some pretty archaic coding style you removed there.

Yes, I agree that using variables for inferred memory is problematic.  In this case it would have been unnecessary because you're using a common clock (so don't actually need separate processes).

PR #253, opened 2025-11-14
axi_lite_slave: fix bvalid could be too early

EmilioPeJu:
When write_ack is 1 constantly, bvalid is asserted before the transfer is
complete(i.e. not AXI compliant) because of its combinatorial nature, this
change fixes the issue by registering bvalid.

Additionally, axi_lite_slave code style was improved.

### DIFF
--- a/common/hdl/axi_lite_slave.vhd
+++ b/common/hdl/axi_lite_slave.vhd
@@ -258,11 +258,7 @@ process (clk_i) begin
                         if vector_and(wstrb_i) then
                             -- Generate write strobe for valid cycle
                             write_strobe_o <= write_strobe;
-                            if write_ack = '1' then
-                                write_state <= WRITE_DONE;
-                            else
-                                write_state <= WRITE_WRITING;
-                            end if;
+                            write_state <= WRITE_WRITING;
                         else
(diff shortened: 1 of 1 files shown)


Araneidae:
This is an odd one.  I don't think your analysis is correct:
1. `bvalid_o` is asserted precisely when `write_state` is `WRITE_DONE`.  Typically this will be implemented as a "one-hot" state machine and `bvalid_o` will be one of those bits.
2. The state `WRITE_DONE` is only entered when the corresponding write acknowledgement `write_ack_i` is seen.

Why do you think this doesn't follow the AXI rules?

EmilioPeJu:
Please allow me to explain in bullet points to see if it makes this clearer:
- The AXI standard says that BVALID shouldn't be asserted before the transfer of address and data
- When the slave has always asserted ack, we transition from IDLE to DONE state and we assert xREADY (which are registers)
- Which means, we have asserted BVALID, but the transfer of data doesn't occur until next tick

EmilioPeJu:
I simplified the change, removed the bypass from IDLE to DONE and tested it, everything is working fine in hardware and simulation also shows the right behavior.

PR #275, opened 2026-05-11
Pbox2 const

glennchid:
Fixing clock skew issue.

Removing need to read MGT pin constraints for panda_sync with the ultra-scale targets. Keeping it the same for the time being for the 7-series targets, and keeping existing pin constraint files for reference, in case they are ever useful.


EmilioPeJu (approved):
Looks good to me

PR #276, opened 2026-05-15
Include opkg-util as a subtree instead of submodule

EmilioPeJu:
Converting opkg-util submodule to subtree serves 2 purposes:
- Prevent people mistakes forgetting to recursive cloning the repo
- Prevent submodule breaking the build when its url goes offline


glennchid (approved):
Seems like a good idea

EmilioPeJu:
Thank you @glennchid

PR #283, opened 2026-06-16
Docs

coretl:


coretl:
Closing during the versioned-docs migration to myst-version-switcher.

The `docs` branch is now published directly as the live `/docs/` site (the cross-repo xrefs point there). While this PR is open, every push to `docs` also fires a `pull_request` event for it, which deploys a redundant `pr-<n>` preview that races the `docs` deploy on the `pages` concurrency group and can clobber `/docs/`.

Closing removes that race. Re-open / recreate this PR when the docs work is ready to actually merge into the default branch (at which point the default branch becomes the live version). Tracked in the docs-migration issue.

